PR #69, opened 2026-06-18
feat(iceberg): real persisted metadata.json + snapshot history (TD-119 Phase 1)

vjsingh1984:
## What & why

Promotes ProximaDB's Iceberg surface from **synthetic, ephemeral** metadata to **real, persisted, spec-faithful** table metadata with a true snapshot history — implementing the user-approved **Phase 1** of TD-119 (design doc `docs/12-design/PROXIMADB_AS_LAKEHOUSE_NATIVE_PG_ENGINE_2026_06_17.md`).

Before: `load_table` rebuilt metadata on every call, `metadata-location` pointed at a non-existent file, snapshots had no parent chain/history, and there was no `refs` map. After: a versioned `metadata.json` is persisted to object storage, the snapshot history is driven by the data manifest log, and `metadata-location` resolves. This makes `LoadTable` trustworthy for external engines **and provides the `refs`/snapshot substrate that unblocks the TD-117 agent-branching chain** — the reason TD-119 was prioritized.

`iceberg-rust` is not a dependency and isn't needed here: Phase 1 is plain serde JSON over the existing metadata types (no Avro).

## What's in scope (each commit tested)

- **`7c3ad2982` — `MetadataCommitter`**: atomic, versioned `v{N}.metadata.json` log over object storage (mirrors `ManifestCommitter`'s `put_if_absent` CAS). 4 tests.
- **`eeb7e8a66` — refs + v2 sequence numbers**: `IcebergSnapshotRef` + a `refs` map (`main` → current snapshot) on `IcebergTableMetadata`, `sequence-number` on snapshots, populated `snapshot_log`. Test asserts spec field names.
- **`313a94e1b` — end-to-end persistence**:
  - `ObjectStoreBridge` gains `latest_metadata_version`/`read_table_metadata`/`commit_table_metadata` (default-unsupported); `IcebergObjectStoreBridge` implements them via `MetadataCommitter`.
  - `IcebergRestService::ensure_table_metadata`: derives the table's object-store base from its catalog location (mirrors the write path's `object_write_base_path`), reads the data manifest log, builds a **parent-chained snapshot chain** (one snapshot per manifest version, stable ids, v2 sequence numbers, `refs.main`), and **idempotently** persists `metadata.json` (only re-commits when the manifest log advanced). **Fail-soft**: persist errors log and still serve inline metadata.
  - `load_table`/`create_table` return a resolvable `metadata-location`; new `GET …/metadata/v{N}.metadata.json` serves the bytes.
  - REST mount builds the warehouse bridge from `file://{data_dir}/warehouse` — the **same root** the materialize/write path uses, so the manifest log aligns.
- **`d73b3de17` — docs**: TD-119 status → Phase 1 landed; `real_iceberg_snapshots` capability `planned`→`partial`.

## Verification
- `proximadb-iceberg-engine` lib: **19/19** (incl. 4 `MetadataCommitter` CAS/round-trip tests).
- `proximadb` `iceberg_rest_service` lib: **7/7**, incl. an **end-to-end** `ensure_table_metadata` test over a `memory://` bridge with a seeded 3-version manifest log → 3 parent-chained snapshots, `refs.main`, idempotent persistence; plus a pure `build_snapshot_chain` test (parent chain, sequence numbers, stable ids).
- `cargo fmt --all --check` clean; capability-matrix + benchmark-evidence validators pass; asciidoctor renders `TECHNICAL_DEBT.adoc`.
- No `.unwrap()/.expect()/panic!` added on production paths; metadata materialization is fail-soft so reads never 500.

## Deferred (Phase 2, tracked under TD-119)
Avro manifest-list (`snap-{id}.avro`) + manifest files with per-file `DataFile` stats — the layer Spark/Trino need to *scan data files* (REST-catalog clients like PyIceberg already get spec-shaped metadata + history inline); exact per-snapshot historical stats; tenant-scoping the Iceberg REST endpoints.


vjsingh1984:
Heads-up (cross-session): `develop`'s **Python SDK Tests** leg is red on `sentence_transformers` failures — filed as #73. Two causes: `ImportError: cannot import name 'SentenceTransformer' from 'sentence_transformers'` (env/namespace) and `_FakeSentenceTransformer.encode() got an unexpected keyword argument 'convert_to_tensor'` (test-fake signature drift) — `tests/unit/test_embedding_providers_v2.py`, `test_atomic_wal.py`, `test_quantization_features.py`. Looks like it came from the embedding/multimodal + arrow58/datafusion54 work, not the API v2 consolidation. The separate URL-validator failure (`test_url_grpc_hostport_kept_asis`) is already fixed + merged via #71. Flagging since you're active in the Python SDK area — details in #73.

PR #97, opened 2026-06-20
feat(rabitq): P3 foundation — block-scoring recall test + segment-migration plan

vjsingh1984:
## What

Lands the **P3 (RaBitQ ~30× quantization) foundation** on top of the merged qa-gate base (#86):

- **`test(rabitq)`** `proximadb-block-format` — `rabitq_block_scoring_preserves_recall_at_k_with_rerank`: proves block-level RaBitQ scoring + f32 rerank preserves recall@k (the quality gate the full wiring must hold).
- **`docs(rabitq)`** scoping doc — the real gap is the **vector segment format** (ProximaDataBlock raw-f32 SST/HELIX/SWIFT), *not* the scan path: RaBitQ codes aren't on the ANN read path yet.
- **`docs(rabitq)`** 7-phase PAX segment-migration plan (mixed-read-safe, so existing f32 segments stay readable during rollout).

## What this is NOT

This does **not** yet wire RaBitQ into the live ANN scan→rerank path — that's the migration plan's scope, deferred to a focused follow-up. This PR is the scoped foundation + recall gate so the wiring can land incrementally against a quality bar.

## Verification
- `cargo test -p proximadb-block-format` → 37/37 pass (incl. the new recall test).
- Rebased cleanly onto develop `2699c665e` (only the 3 unique commits; qa-gate base dropped).

### DIFF
--- a/crates/storage/proximadb-block-format/src/reader.rs
+++ b/crates/storage/proximadb-block-format/src/reader.rs
@@ -981,6 +981,107 @@ mod tests {
         assert!(cos > 0.3, "reconstruction cosine {cos} too low");
     }
 
+    /// RaBitQ recall@k with f32 rerank, through the on-disk stripe scoring path — the
+    /// contract the ANN search executor will rely on: rank candidates by the binary
+    /// estimator over decoded codes, then rerank the top candidates against full f32.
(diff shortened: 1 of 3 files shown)


vjsingh1984:
Folding P3 into a single consolidated PR with the P2 CDC surface and P4 freshness work — one CI run instead of three (see new combined PR).

PR #118, opened 2026-06-21
feat(p3): constant-column encoding for single-valued Tenant/Identity strings

vjsingh1984:
## P3: constant-column encoding for single-valued Tenant/Identity strings

A single-tenant PAX segment stored `tenant_id` as a **dictionary** (1 entry + N per-row codes ≈ N bytes) even though every row is identical. This makes that column **O(1)** when it's constant.

### Why the per-row column stays (and why this is the right fix)
Tenant isolation is already **structural** — the `DrPathBuilder` path (`data/{tenant}/{namespace}/…`) plus the block-header `tenant_id_hash` RLS skip. The per-row `tenant_id` column exists only for (1) canonical-record fidelity (authority = WAL + `ProximaRecord`), (2) row-level RLS in shared (header-hash 0) blocks, and (3) fail-closed defense-in-depth. So we keep it — but stop paying ~N bytes for a constant.

### Changes
- **Writer**: a constant string column (`ColumnRole::Tenant`/`Identity`, `distinct == 1`, no nulls) is emitted as `[value_len][value]` under the **existing `RunLength` wire scheme** (marker `0x62` — its documented intent is "constant/near-constant"). No new `ProximaScheme` variant → no match fan-out. `select_str_scheme` gains the constant branch; `encode_str_with_scheme` dispatches `RunLength`.
- **Reader**: `decode_str_with_encoding` handles the `RunLength` marker, reconstructing N copies. **Always decodable** → any reader carrying this change is forward-compatible regardless of writer flag state.
- **Gating** (storage-format-migration mandate): default **OFF** via the `with_constant_str` builder (mirrors Phase D's `with_quant`), default from `PROXIMADB_PAX_CONSTANT_STR`. `PaxSegmentWriter` inherits it through `PaxBlockWriter::new`, so production stays on dictionary until readers are deployed, then the flag flips on. **No default behaviour change.**

### Verification
- `proximadb-block-format` leaf: **39/39** (new `constant_str_tenant_column_round_trips_and_shrinks` — round-trips all N rows, `RunLength`-marked, strictly smaller than the dictionary the default flag-off writer still emits).
- `cargo fmt --all` (only touched files), leaf + root `clippy` clean, `CARGO_INCREMENTAL=0`.

### DIFF
--- a/crates/storage/proximadb-block-format/src/reader.rs
+++ b/crates/storage/proximadb-block-format/src/reader.rs
@@ -498,8 +498,26 @@ fn decode_str_with_encoding(
     match scheme {
         ProximaScheme::Raw => decode_raw_str_col(data, count),
         ProximaScheme::Dictionary => decode_dictionary_str_col(data, count),
+        // Constant-column (single stored value → N copies). Always decodable so a
+        // reader is forward-compatible regardless of the writer's flag state.
+        ProximaScheme::RunLength => decode_constant_str_col(data, count),
         other => bail!("unsupported PAX string encoding: {}", other.name()),
     }
+}
+
+/// Decode a constant string stripe (`[value_len u32][value bytes]`) into `count`
+/// copies of the single value — the inverse of the writer's `encode_str_constant_col`.
+fn decode_constant_str_col(data: &[u8], count: usize) -> Result<Vec<Option<String>>> {
+    if data.len() < 4 {
+        bail!("constant string stripe missing value length");
+    }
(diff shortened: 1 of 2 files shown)


vjsingh1984:
Superseded by the catalog-resolution approach: since every segment is path-isolated to one tenant (data/{tenant_id}/{namespace_id}/...), tenant_id is redundant in the segment. Instead of constant-encoding the column we'll drop it and stamp record.tenant_id from the (tenant, namespace) catalog/path context on read. Closing in favor of that cleaner design.

PR #145, opened 2026-06-21
docs(design): correct decomposition first-slice to kv (formats is engine-bound)

vjsingh1984:
Follow-up to the decomposition plan (#143). A file-level check disproved the formats-first recommendation: `formats/adapters.rs` binds all six engines (`crate::storage::engines::*`, 12 refs), so formats cannot move ahead of the engines. This corrects the plan to the verified clean leaves (kv 96 LOC zero-deps; transaction; encryption; schema; tenant) and sets Slice A to `proximadb-storage-kv` as the smallest provable inch.


vjsingh1984:
Superseded/overtaken by the actual root-crate-decomposition work (#329 landed; #339/#348/#357/#359 in flight). This docs first-slice plan is stale and conflicting; closing.

PR #226, opened 2026-06-22
feat(catalog): pluggable CatalogSnapshotStore + object-store snapshot persistence (system-catalog 5c)

vjsingh1984:
## Summary

**Phase 5, slice 3** of the two-tier operator/account model (builds on #213, #222). Makes the system catalog's **snapshot blob persist to real object storage** (`s3`/`gs`/`az`/`memory`) under the tenant/operator `DrPathBuilder` prefix — replacing the all-local snapshot path — **without touching the Phase-3 crash-consistency ordering or the local WAL**.

## Changes
- New module `services::catalog_snapshot_store` — **`CatalogSnapshotStore`** trait (`read → Option`, atomic `write_atomic`, `describe`) with two impls:
  - **`LocalSnapshotStore`** — temp → fsync → atomic rename (the prior inline `write_atomic`, moved out verbatim; the default).
  - **`ObjectStoreSnapshotStore`** — single whole-object PUT/GET via `ProximaObjectStore` (atomic at the object level; `NotFound → None` for the empty-catalog boot case). Works over `memory://` and, behind the crate's cloud features, `s3/gs/az`.
- **`SystemCatalog`** snapshot config now holds an `Arc<dyn CatalogSnapshotStore>` instead of a raw path. `open` keeps the local default (behaviour-identical); new **`open_with_snapshot_store`** injects any store; `checkpoint` writes through the store. The "snapshot decode failure is fatal" invariant is preserved.
- Root crate gains a `proximadb-object-store` dependency (leaf crate, no cycle).

## Scope boundary
The **per-DDL WAL stays local**; object-store-*native* WAL durability is **Phase 6** (its delta-object log is the multi-pod CAS substrate — building it here would duplicate it). Boot cutover for object-store metadata URLs + removal of `NativeCatalog::parse_storage_url`'s temp-cache fake is the next slice (**5d**).

## Verification
- `cargo clippy --lib --bins -- -D warnings` — clean
- `cargo test -p proximadb --lib -- catalog_snapshot_store system_catalog` — **18/18**, incl. a `memory://` round-trip and a catalog **reopen restoring from an object-store snapshot + local WAL tail**
- `check_tenant_path_guard.py` — green
- Rebased onto develop to pick up the #219/#221 `quantization_engine` path fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6543,6 +6543,7 @@ dependencies = [
  "proximadb-kernel",
  "proximadb-multimodel-plan",
  "proximadb-multimodel-query",
+ "proximadb-object-store",
  "proximadb-observability",
  "proximadb-observability-query",
  "proximadb-pipeline-operator",
@@ -6975,6 +6976,7 @@ dependencies = [
  "num_cpus",
  "proximadb-config",
  "proximadb-hardware",
+ "raw-cpuid",
  "serde",
  "tracing",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -299,6 +299,7 @@ proximadb-runtime-common = { path = "crates/horizontal/proximadb-runtime-common"
 proximadb-security = { path = "crates/horizontal/proximadb-security" }
 proximadb-storage-common = { path = "crates/storage/proximadb-storage-common" }
 proximadb-block-format = { path = "crates/storage/proximadb-block-format" }
+proximadb-object-store = { path = "crates/storage/proximadb-object-store" }
 proximadb-iceberg-engine = { path = "crates/storage/proximadb-iceberg-engine" }
 proximadb-storage-kv = { path = "crates/storage/proximadb-storage-kv" }
 proximadb-storage-optimization = { path = "crates/storage/proximadb-storage-optimization" }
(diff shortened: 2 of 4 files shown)


vjsingh1984:
Heads-up: this broke develop's `proximadb` lib compile — `src/services/catalog_snapshot_store.rs` is **not committed** (it's in no git ref), but `src/services/mod.rs:176` declares `pub mod catalog_snapshot_store;` and `src/services/system_catalog.rs` imports `CatalogSnapshotStore`/`LocalSnapshotStore`/`ObjectStoreSnapshotStore` from it:

```
error[E0583]: file not found for module `catalog_snapshot_store`
  --> src/services/mod.rs:176
```

Cause is the known macOS `.gitignore` trap — line 414 `*snapshot*` silently ignores `catalog_snapshot_store.rs`, so it never got committed (your local build works because the file is on disk). `git cat-file -e origin/develop:src/services/catalog_snapshot_store.rs` → does not exist; `git check-ignore` confirms `.gitignore:414 *snapshot*` matches it.

**Fix (needs the file, which only you have):**
1. Add a negation to `.gitignore` next to the existing ones (`!src/**/snapshot.rs`, `!src/services/snapshot/`):
   `!src/services/catalog_snapshot_store.rs`
2. `git add -f src/services/catalog_snapshot_store.rs` and commit.

It slipped through CI because compile builds appear non-gating. develop's lib is red for every checkout right now. Can you push the missing file?

PR #254, opened 2026-06-24
feat(pgwire): ON CONFLICT DO UPDATE / DO NOTHING upsert (TD-129)

vjsingh1984:
Idempotent re-push of a code-graph file batch must update a symbol row in
place, not duplicate it. The parser/AST/routing already carried the ON
CONFLICT clause; execute_upsert ignored it (unconditional upsert-by-oid).

**What changed**
- Wire the conflict clause through the DML boundary:
  - Conflict target must reference the single-column PRIMARY KEY (the record
    oid is PK-derived, so a re-push of the same key resolves to the same oid).
  - Per-row point-probe → DO UPDATE (reusing the UPDATE applier with
    excluded.* pre-resolved to the proposed INSERT literal) / DO NOTHING /
    insert.
  - Within-statement duplicate conflict keys rejected (Postgres
    "cannot affect row a second time").
  - Postgres-accurate affected count; only inserts bump row-count stats.
- Reuses build_updated_proxima_record + get_by_key; no new decoder.
- e2e test proves re-upsert leaves one row updated, DO NOTHING preserves it,
  new key inserts.

**Verification**
- `cargo nextest run pgwire_upsert_on_conflict_e2e` — green (1/1 passed)
- `cargo clippy -p proximadb --lib` — 1 style suggestion, no panic/unwrap hits
- fmt check passes on touched files

Refs TD-129.

### DIFF
--- a/docs/10-quality/TECHNICAL_DEBT.adoc
+++ b/docs/10-quality/TECHNICAL_DEBT.adoc
@@ -376,10 +376,10 @@ a| *RESOLVED Mar 2026*: **Cypher query language fully implemented**. Complete Op
 | pgwire — `ON CONFLICT DO UPDATE`
 | P2
 | MED
-| *Open*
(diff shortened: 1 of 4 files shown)


vjsingh1984:
Obsoleted by #272 (rebased onto develop with proto drift baseline fix)

PR #402, opened 2026-06-26
fix(test): switch volcano streaming tests to current_thread runtime (hang flake)

vjsingh1984:
## Problem

The `native_volcano_stream_*` tests intermittently hung in CI until the 240s nextest TMT timeout (3 retries × 240s ≈ 12min wasted per CI run). The code path is sequential (`stream::try_unfold` + `next_row`, no spawn/channel/task) — root cause unclear, likely CI scheduling.

## Previous attempt (reverted)

`#[tokio::test(flavor = "current_thread")]` — **counterproductive**: the single-threaded runtime's drop blocks on the hung test's pending tasks → process-level hang → the job took **1h20m** (worse than the original ~1hr). Reverted.

## Fix

Wrap `execute_physical_stream` in `tokio::time::timeout(Duration::from_secs(10))`. If the streaming path hangs, the test **fails fast** (10s) instead of blocking for 240s × 3 retries. Preserves `#[tokio::test]` (multi-threaded, original behavior). Root cause TBD — needs a deterministic repro.

## Verification

The CI run will show whether `Rust Tests (unit)` completes faster (~30-40min, no hang) vs the previous ~1hr+. No production code changed.

### DIFF
--- a/src/query/execution/engine.rs
+++ b/src/query/execution/engine.rs
@@ -387,6 +387,7 @@ mod tests {
     use proximadb_relational_planner::PhysicalPlan;
     use proximadb_relational_reader::RelationalReader;
     use proximadb_relational_types::{ColumnInfo, Expr};
+    use std::time::Duration;
 
     struct EmptyReaderFactory;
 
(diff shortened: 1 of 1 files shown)


vjsingh1984:
Heads up — opened #409 to consolidate this. I reproduced the flake locally and the hang is in **`rows.next()`** (the drain), not the `execute_physical_stream(...)` open that this branch wraps — so the open-only wrapper misses it (and your branch doesn't cover `enforces_row_limit`, which is the test I saw hang). #409 wraps the whole body of all 4 streaming tests with a 10s timeout, keeping your fail-fast intent. Your `current_thread` finding (drops block) is why #409 stays on the default multi-thread runtime. Happy to fold in anything else.

vjsingh1984:
Closing — superseded by #409 (`fix(test): bound all native_volcano_stream tests with a 10s timeout`), which landed the whole-body `tokio::time::timeout` wrapper (the right fix — covers `rows.next()` too, which my partial wrapper on `execute_physical_stream` didn't). Also #400 re-required Rust Tests (unit) + cleared the unit-test rot so the gate passes. The streaming-hang flake is resolved on develop.

PR #403, opened 2026-06-26
fix(ci): make native_volcano_stream flake deterministic (10s timeout)

vjsingh1984:
Fixes the known streaming-hang flake (CLAUDE.md mandate #11) that was blocking every PR's 'Rust Tests (unit)' check by hanging the CI runner indefinitely.

Wraps the test body in `tokio::time::timeout(10s)` — it now **fails fast** with a clear error if the Limit executor race triggers, instead of hanging the runner for hours. The test still passes normally (the race is intermittent).

### DIFF
--- a/src/query/execution/engine.rs
+++ b/src/query/execution/engine.rs
@@ -635,25 +635,33 @@ mod tests {
 
     #[tokio::test]
     async fn native_volcano_stream_truncates_without_error() {
-        let controls = ExecutionControls {
-            max_rows: Some(1),
-            row_limit_mode: RowLimitMode::Truncate,
-            ..Default::default()
-        };
-        let result = NativeVolcanoEngine::execute_physical_stream(
-            values_plan(),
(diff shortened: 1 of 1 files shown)


vjsingh1984:
Superseded by #409. I reproduced the flake locally (`cargo nextest run -p proximadb --lib native_volcano_stream`): the test that actually hangs is **`native_volcano_stream_enforces_row_limit`** (180s on try 1, 30ms on try 2), not `truncates_without_error` that this PR wraps. #409 takes your whole-body timeout approach (correct — guards the `rows.next()` drain too) and applies it to all 4 streaming tests. Suggest closing this in favor of #409.

vjsingh1984:
Closing — superseded by #409, which merged (9f848f37b). #409 applies the same whole-body timeout to all 4 native_volcano_stream tests (this PR only covered truncates_without_error).

PR #436, opened 2026-06-27
docs(adr): ADR-033 catalog-object surrogate identity — immutable u64 oid, name as mutable attribute

vjsingh1984:
## Summary

Records **ADR-033**: catalog-object identity (accounts, namespaces, tables/collections, columns, indexes) is an **immutable, system-generated `u64` `object_oid`**, with the **name as a mutable attribute**. Refines `CATALOG_OBJECT_MODEL_2026_06_21` (today's `{namespace_id=ns_<uuid>, tenant_id, name, object_id}` string identity).

## Why

A catalog object's identity is the *object*, not its *name* — a renamed table/namespace/account is the same object.

- **Rename survival**: every internal reference + the DrPath key off the oid, so rename is an O(1) metadata update — never a data move or reference rewrite (today's name/uuid-in-path makes rename risk O(data)).
- **Performance**: fixed-width `u64` lookup/join/FK beats the current ~40-char UUID-string on the catalog's hottest path (table → namespace → tenant resolution on every query) — faster compare, denser index, less storage.

## Decisions (D1–D6)

- **D1/D2** — immutable `u64` surrogate; name is a mutable attribute; all references use the oid; name→oid resolved once at parse.
- **D3** — `u64` (not `u32` — Postgres OID wraparound at multi-tenant scale); system-wide monotonic from the central catalog sequence (coordination is free at DDL rates); never reused (dangling-detectable).
- **D4** — reserved low oid range for system objects (`ns_default` → fixed oid).
- **D5** — stored as an indexed `ProximaType::UInt64` column (reuse ADR-024, additive; logical identity ≠ physical record handle).
- **D6** — DrPath keys off the oid (`data/{tenant_id}/{namespace_oid}/{table_oid}/`) → rename never moves data.

## Scope

Catalog objects (metadata) only. The data-record oid (`ProximaRecord.oid: String` / ULID) is a separate, settled concern (high-rate, distributed, coordination-free) — deliberately not conflated.

## Status

**Proposed** — decision only. Implementation (oid sequence + name→oid resolution + the DrPath migration behind a compat shim) is **TD-174**, sequenced (not big-bang). Reuses `ProximaType::UInt64` (ADR-024). Rejected alternatives (name-as-identity, UUID-string, `u32`, random u64, ULID-for-catalog, per-tenant oid space) are documented in the ADR.

Refs: ADR-024, ADR-027 (DrPath), `CATALOG_OBJECT_MODEL_2026_06_21`, AnvaiOps ADR-0021; TD-174.

### DIFF
--- a/docs/12-design/CATALOG_OID_MIGRATION_2026_06_27.adoc
+++ b/docs/12-design/CATALOG_OID_MIGRATION_2026_06_27.adoc
@@ -1,0 +1,69 @@
+// Copyright (C) 2025 ProximaDB
+// SPDX-License-Identifier: Apache-2.0
+= Catalog-Object OID Migration Plan (TD-174 — ADR-033 execution)
+:toc: left
+:toclevels: 3
+:icons: font
+:source-highlighter: rouge
+:last-updated: 2026-06-27
+
+Paired ADR: link:adr/ADR-033-catalog-object-surrogate-identity.adoc[ADR-033] (the decision). Tracker: TD-174.
+
+== Goal
+
+Give every catalog object (account / namespace / table-collection / column / index) an immutable, system-generated `u64` `object_oid` (ADR-033); make every internal reference **and the storage path** key off it; and demote the name to a mutable attribute — so **rename is an O(1) metadata update, never a data move**, and catalog joins/FKs/lookups are fixed-width integer ops. Phased, durable-safe, behind a compat shim — never big-bang.
+
+== Invariants (a test enforces each, at every phase that touches it)
+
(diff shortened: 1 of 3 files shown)


vjsingh1984:
Superseded by the canonical statistics line (ADR-037/038, TD-174/175/176) and the object-id decision (ADR-031 stable-object-id + the adr031-033-reconcile branch). This PR's ADR-033 duplicates the merged ADR-033 (primary-storage-format-strategy) and would hard-fail the new #441 td-adr-unique guard on rebase. Closing. Salvageable for the object-id reconcile (on this branch's commit 59d0dac7, kept un-deleted): CATALOG_OID_MIGRATION_2026_06_27.adoc (the no-data-move DrPath indirection + phased oid migration) and the ADR-024 identity addendum (record String/ULID vs catalog u64 vs pgwire_oid).

PR #439, opened 2026-06-27
ci(test): jobs=2 compile + swapoff before run (split fix for #433 regression)

vjsingh1984:
## Summary

Re-applies the `jobs=2` compile speedup from #433 but fixes its regression by **splitting build from run and dropping the swapfile in between**, so each phase gets the right memory posture.

### Why #433 regressed
The unit job is compile-bound — measured **~42 min compile + ~15 min run** (~57 min, jobs=1). #433 sped the compile (~24 min) behind a 12 G swapfile to survive the link-time OOM, but **leaving that disk-backed swap up during the run** let the 8300-test phase thrash **~3× slower** (902s → 2661s), a net regression (~57 → ~69 min).

### The fix
```bash
CARGO_BUILD_JOBS=2 cargo nextest run --lib --profile unit --no-run   # compile (~24m), swap covers OOM
sudo swapoff /swapfile-ci || true                                    # remove run-phase thrash
cargo nextest run --lib --profile unit --test-threads=2             # run already-built, RAM-only (~15m)
```
Projected **~42–45 min** vs the 57-min baseline. Swap is now a **compile-only** margin.

## Validation
`.github/workflows/ci.yml` re-added to the `rust:` paths-filter, so **this PR self-validates on its own unit run**. Watch for:
- compile step ≈ 24 min (jobs=2 working), no exit-143 OOM,
- `swapoff` log line confirming swap removed before tests,
- nextest `Summary` back to ~900s (run no longer thrashing),
- total unit job ≈ 42–45 min.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,10 @@ jobs:
               # workspace-wide, so a bump must re-run the Rust suite.
               - 'rust-toolchain*'
               - '.cargo/**'
+              # The CI workflow itself defines HOW the Rust suite compiles/runs
+              # (build-jobs, test-threads, nextest profile, swap headroom); a change
+              # here must re-run the suite so the change self-validates on its own PR.
+              - '.github/workflows/ci.yml'
             # `compile` gates the rust-compile hard gate. It is a deliberate
             # SUPERSET of every path that can affect `cargo check --workspace`:
             # the root `src/**`, every workspace member under `crates/**` and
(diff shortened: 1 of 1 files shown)


vjsingh1984:
Closing without merge. The build/run split worked mechanically — compile dropped to ~24min (jobs=2), swapoff confirmed (14G→3G before tests), no OOM — and it recovered ground vs #433 (69→61min). But the **test-run phase remains the real bottleneck** (~36min vs the ~15min baseline sample; swapoff only partially mitigated it), so the total (~61min) is not a clear win over the ~57min jobs=1 baseline and sits within the noisy 50–81min range. Not worth the swapfile+split complexity. Staying on the simple jobs=1 baseline.

The genuine lever is the **run** phase, which only nextest **archive + shard** (compile once → run partitions in parallel) actually attacks — deferred unless CI wall-clock becomes a priority.

PR #682, opened 2026-07-05
feat(olap): TD-OLAP-6 sort-on-materialize — static pruning 0% → 19.2% measured

vjsingh1984:
## Summary

The final link of the pruning chain (v1→v2→v3 evidence arc). MATERIALIZE now **clusters the snapshot before publication** — explicit `cluster_key` table property, else the first DATE/TIMESTAMP column heuristic; NULL-keyed rows last; one sort at publication; reader-neutral (mixed-read-safe).

### Measured (evidence v3, checked in)
- **TPC-H static split skip: 0% → 81/422 (19.2%)** on identical queries/data. Q14/Q15 prune 14/16 row groups; **Q6 prunes 12/15 and runs in 50 ms** beside multi-second peers.
- TPC-DS static 0% *as expected* (its date predicates live on `date_dim`; the fact-table case belongs to the runtime join filter).
- **`PROXIMADB_DF_RUNTIME_FILTER_PRUNE` measured zero incremental pruning** (identical 81/422, no per-query delta) — stays default-off per the promote-on-measured-win mandate; TD-OLAP-3 now carries the follow-up: instrument the dynamic-filter absorption path (does DF's Post-phase pushdown reach `ProximaScanExec` on our plan shapes?).
- Conformance unchanged: TPC-H 22/22, TPC-DS 16/16, delta-merge green; clippy/fmt clean; server binary builds.

### Tests
- `materialize_publishes_clustered_row_groups`: asserts monotone, near-disjoint per-row-group date windows (the regression mode was every group spanning `[8766, 9066]`); windows now e.g. `[8766,8826] [8826,8888] [8888,8950] [8951,9014] [9014,9066]`.
- Cluster-key resolution unit test (property precedence, bogus-property fallback, NULL-last ordering).

### Docs
`TPC_PERF_GATE_EVIDENCE_V3_2026_07_05.adoc` (results + decisions), TD-OLAP-6 status (slice 1 landed; remaining: `CLUSTER BY` DDL, Iceberg sort_orders), TD-OLAP-3 status, BENCHMARK_EVIDENCE note.

Note: this branch was replayed from a fresh worktree after a concurrent-session disk cleanup removed the in-progress one; the transiently 'failed' gated evidence pass is explained by that removal (clean repro passes 38/38).

Refs: ADR-052 (P2a arc), TD-OLAP-6, TD-OLAP-3, TD-OLAP-4.


vjsingh1984:
Pushed two additional commits from the absorption-path diagnosis (task ran while CI was green on the first commit):

1. **Defer runtime-filter pruning to first stream poll** — instrumentation showed the full pushdown chain works (Post-phase delivery, downcast, absorption) but `snapshot()` at `execute()` time always saw the `lit(true)` placeholder: DataFusion constructs the probe stream before the build side populates the filter.
2. **Per-split late evaluation** (`DataSourceExec` semantics) — partitioned hash joins poll build and probe concurrently, so even first-poll can race; each split now checks the *current* snapshot immediately before its read. Conservative: a split evaluated too early is read, never wrongly skipped.

Both are correctness improvements to the gated path (conformance green with the gate on). Measured result unchanged — zero incremental pruning — so the gate **stays default-off**; evidence v3 and TD-OLAP-3 now record the precise remaining suspects (does the join's rebuilt node survive our optimizer pipeline; is the post-build expr the hash-lookup form the translator skips). The PR's headline remains sort-on-materialize: **0% → 19.2% static split skip.**

PR #872, opened 2026-07-11
feat(olap): ADR-059 — DuckDbLocalEngine in-process engine + ClickBench/TPC harness + catalog location fix

vjsingh1984:
## What
Implements ADR-059 (DuckDB-Local plugin OLAP engine) + rewires both perf-ledger harnesses to use it in-process (io-traced, apples-to-apples) + fixes the catalog location model (first-principles, Hive/Unity/Polaris-style).

## 1. DuckDbLocalEngine + dispatch (feature-gated `duckdb`, default OFF)
- In-process DuckDB behind the `ExecutionEngine` seam. Registers parquet as `CREATE VIEW ... read_parquet('{location}')` (same data), executes on the blocking pool, converts Arrow via the shared `record_batches_to_pipeline_result` bridge, records engine-reported io-trace (`compute_ms`).
- `DuckDbCompat` dispatch arm in `execute_sql_with_backend`.
- Arrow versions unify at 58.3.0 (datafusion 54.0 + duckdb 1.10501).

## 2. ClickBench harness — in-process DuckDB (io-traced)
Replaces the CLI subprocess with `execute_sql_with_backend(DuckDbCompat, ...)` in `io_trace::instrument` → `drain_capture` → `QueryRecord` with `compute_ms` + real `rows`. CLI kept as `#[cfg(not(feature="duckdb"))]` fallback.

## 3. TPC harness — in-process DuckDB (apples-to-apples, same parquet)
Replaces the INSERT-based CLI baseline with the in-process engine reading the SAME materialized parquet DF reads. Queries `xcatalog.table_routing` for per-table parquet locations (post-MATERIALIZE), filtered to the benchmark's tables. io-traced via `io_trace::instrument` + `drain_capture`.

## 4. Catalog location model (first-principles — Hive/Unity/Polaris)
- **`location` exposed** in `xcatalog.table_routing` (the table's data dir where parquet is immediate).
- **`primary_layout` prefers Parquet** (the materialized layout, not the stale original).
- **Materialize sets `location = {root}/{prefix}/data`** (where parquet is immediate children, not the table base dir with a `data/` subpath).
- **DF reader mixed-read-safe** (scan `{location}` immediate + fallback to `{location}/data/` for old tables).
- **`cataloged_tables` dedup prefers the Parquet catalog** (fixes the stale-location bug — `set_storage_layouts` updates the internal `proximadb.` catalog, but the dedup was keeping the stale `public.` form).
- **`metadata_url` per-tempdir** in the test (fixes the stale-catalog-across-runs bug — the default `file://./metadata` persists across test runs).

## 5. Native grouped-agg env gate (B2)
`PROXIMADB_NATIVE_ALLOW_GROUPED_AGG=1` opts in to native grouped aggregates (the caller ensures low cardinality, e.g., Q1's 8-group GROUP BY at SF=0.01).

## Verification
- `cargo clippy --lib --bins -- -D warnings` clean (default build).
- `cargo clippy --lib --bins --features duckdb -- -D warnings` clean (libduckdb 1.10501 builds via `bundled`).
- SF=0.001: DuckDB ok 22/22 (tpch) + 63/63 (tpcds); DF 22/22 + 63/63 (conformance unchanged); DuckDB median 9ms/13ms vs DF 63ms/75ms.
- Rebased onto latest develop (10 new commits — no regression).

## Non-goals (follow-ons)
- **A2 (HLL during materialize)** — the proper TD-OLAP-2 NDV fix (the `num_rows`-as-NDV proxy was reverted — it caused a bad DF plan at SF=0.01 due to overestimated NDV for low-cardinality columns).
- **Native morsel scheduling** (B1) — enable `PROXIMADB_NATIVE_MORSEL` + verify parallel scan.
- **Native join + multi-table parquet** (B3) — extend `try_native_over_parquet` for >1 table.
- **Default-location resolution + partitioned layouts** — follow-ons to the catalog location model.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,6 +51,17 @@ dependencies = [
 
 [[package]]
 name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
+dependencies = [
+ "getrandom 0.2.17",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
+name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
@@ -180,8 +191,8 @@ dependencies = [
  "regex-lite",
  "serde",
  "serde_json",
- "strum",
- "strum_macros",
+ "strum 0.25.0",
+ "strum_macros 0.25.3",
  "thiserror 1.0.69",
  "typed-builder",
  "uuid",
@@ -206,6 +217,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arbitrary"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
+dependencies = [
+ "derive_arbitrary",
+]
+
+[[package]]
 name = "arc-swap"
 version = "1.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -225,6 +245,12 @@ name = "arraydeque"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "arrow"
@@ -268,7 +294,7 @@ version = "58.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cfd33d3e92f207444098c75b42de99d329562be0cf686b307b097cc52b4e999e"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "arrow-buffer",
  "arrow-data",
  "arrow-schema",
@@ -467,7 +493,7 @@ version = "58.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8cd065c54172ac787cf3f2f8d4107e0d3fdc26edba76fdf4f4cc170258942222"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "arrow-array",
  "arrow-buffer",
  "arrow-data",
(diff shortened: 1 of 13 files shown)


vjsingh1984 (commented):
## Fresh-context adversarial review — VERDICT: NEEDS FIXES (1 blocker, else clean)

Reviewed the full diff + ADR-059 against `origin/develop`. The design is sound; the only thing standing between this and green is a two-line test-fixture update.

### 🔴 Blocker — the red "Rust Tests (unit)" (trivial fix)
Root cause is **not** DuckDB and **not** environmental — it's a location-semantics change whose guard tests weren't updated in the same commit. `dml/mod.rs:1509` now publishes the catalog location as `{root}/{prefix}/data` (ADR-059 "location = dir where parquet is immediate"), but two tests still assert the old path (these compile under **default features**, so the `duckdb` feature is OFF and the new engine isn't even involved):

- `src/services/dml/tests.rs:2640` — `location.ends_with("/inv")` → **`.ends_with("/inv/data")`**
- `src/services/dml/tests.rs:2913` — `location.ends_with("/inv")` → **`.ends_with("/inv/data")`**

(The `!contains("default_tenant")` / `starts_with(".../ns_")` assertions stay as-is.) Fix these two and CI goes green.

### ✅ Clean (verified)
- **Dependency:** `duckdb`/`libduckdb-sys` are `optional=true`, feature `duckdb` **default-OFF**, not in any default/transitive set; Windows-ratchet green (default features skip it); MIT-licensed; `bundled` (vendored). Correct for a heavy native dep.
- **ADR-039 seam:** no `duckdb::` leakage past `duckdb_engine.rs`; implements `ExecutionEngine`, dispatched via the pre-existing `ComputeBackend::DuckDbCompat` arm; router doesn't auto-select it yet (harness-only wiring, consistent with ADR-059 "Proposed").
- **Thread-safety:** the non-`Sync` DuckDB `Connection` is fully synchronous inside `spawn_blocking` — never crosses `.await`. Correct.
- **Panic policy:** `duckdb_engine.rs` is all `Result`+`?`; `.expect()`s are test-only.
- **Harness rewire:** additive (native/DataFusion routes + ratchets untouched); mixed-read-safe reader fallback; and it fixes a real bug (per-tempdir `metadata_url` vs the stale fixed `file://./metadata`).

### Nits (non-blocking)
- The DuckDB ledger row is **compute_ms-only** (no `bytes_read` — DuckDB's C++ reader has no ObjectStore hook), so DuckDB-vs-DataFusion is a compute-term-only comparison. Worth a note where the ledger is read.
- ADR-059 is still **Proposed** (decider sign-off pending) — harness-only wiring is low-risk to land ahead of sign-off, but flag it before this becomes a permanent engine.

Nice work on the seam discipline and the `spawn_blocking` handling — those are the two easiest things to get wrong with an embedded native engine, and both are right.

vjsingh1984:
Re: the adversarial review — applied the location-assertion fix (`.ends_with("/inv")` → `.ends_with("/inv/data")` at tests.rs:2642 and :2913), but that alone was **not sufficient**: `materialize_table_writes_parquet_and_flips_catalog_layout` then failed at the read-back (line 2652) with `…/inv/data/data/part-0.parquet not found` — the doubled `/data`. The test derived its read path as `{prefix}/data/part-0.parquet`, but under the ADR-059 location semantics the prefix already includes `/data`, so the parquet is an immediate child: fixed to `{prefix}/part-0.parquet`. Both `materialize_*` tests verified passing locally (default features). Pushed in 95308dc7f.

PR #890, opened 2026-07-12
perf(olap): TD-OLAP-17 — skip changed_oids_since scan via per-collection high-water

vjsingh1984:
## What

TD-OLAP-17 fix: skip the per-table-per-query `changed_oids_since` WAL scan (the real DF↔DuckDB gap) via a per-collection high-water in the WAL appender.

## Why (the measured reframe — TD-OLAP-17)

The DF↔DuckDB gap was **not** the join. It was `read_changes_since` calling `wal_appender.read_all_entries()` (O(WAL size)) + filtering by `collection_id`, **per table per query**, in the default-ON `olap_delta_merge` path. Measured: `open_ms ≈ wall_ms` at SF=0.01 + SF=0.1; the join execution itself is 8–23 ms (<1% of wall) even at SF=0.1. Disabling the scan (`PROXIMADB_OLAP_DELTA_MERGE=0`) took DF from **5762 ms → 12 ms** median at SF=0.1 — faster than DuckDB (14 ms). The "single-threaded join build / 100–700× gap" was a misattribution.

## How

- New `TableWalAppender::collection_max_sequence_for(collection_id) -> Option<u64>` — the highest sequence_number for that collection, bumped **atomically on append** (under `append_lock`, after `sync_data` so a reader observing it is guaranteed the entry is durable).
- `read_changes_since` / `read_changes_since_scoped` fast-path: if the collection's high-water hasn't advanced past `since_lsn` → return empty **without** the scan.
- The shared canonical WAL is tagged per entry, so the watermark is **per-collection** (a global max would be ineffective — one table's write keeps it > every other table's snapshot_lsn).

## Properties

- **Strictly an improvement**: worst case = scan = original behavior; never regresses.
- **Correctness**: the high-water is bumped under `append_lock` after `sync_data`, so it's race-free vs concurrent reads (a reader seeing `max > since_lsn` is guaranteed the entry is durable → a scan finds it).
- **Restart**: the high-water is in-memory; after a restart the first read per collection scans (original behavior) until the next write repopulates it. A recovery-rebuild is a follow-on (never regresses — read-only-after-restart = original scan).

## Bonus: latent bug fix

`MemoryTableWalAppender` did not override `read_all_entries` (trait default = empty), so the CDC feed + OLAP delta merge were **silently empty** on the `opt_config=None` boot path. Now overridden to return the in-memory entries.

## Test

- New unit test `read_changes_since_skips_scan_via_collection_high_water` — proves the fast-path (high-water ≤ since_lsn → empty, no scan) + the scan path (high-water > since_lsn → entries) + the unknown-collection path.
- The 26-test `record_store` + `canonical_wal` suite stays green (no regressions).

## Measured (SF=0.01, default config — to be added post-build)

[perf verification pending the release build; expected: DF median drops from ~604 ms to near DuckDB's ~10 ms without `PROXIMADB_OLAP_DELTA_MERGE=0`]

Refs: TD-OLAP-17, ADR-025

### DIFF
--- a/src/services/canonical_wal.rs
+++ b/src/services/canonical_wal.rs
@@ -35,6 +35,25 @@ pub struct FramedTableWalAppender {
     path: PathBuf,
     next_sequence: AtomicU64,
     append_lock: Mutex<()>,
+    /// TD-OLAP-17: per-collection high-water sequence_number, bumped atomically
+    /// under `append_lock` on every append. Lets `read_changes_since` skip the
+    /// O(WAL) scan when a collection hasn't advanced past a reader's `since_lsn`
+    /// (the DF table-OPEN floor). Sync lock (parking_lot) so the sync trait method
+    /// `collection_max_sequence_for` can read it without `.await`. Not rebuilt on
+    /// recovery → after a restart the first read per collection scans (the original
+    /// behavior) until the next write repopulates it; a recovery rebuild is a
+    /// follow-on. Never regresses (worst case = scan).
+    collection_max: parking_lot::Mutex<std::collections::HashMap<String, u64>>,
+}
+
+/// Extract the `collection_id` a canonical operation targets, if any (the
+/// shared WAL tags every entry; non-record ops carry none).
+fn collection_id_of(op: &CanonicalOperation) -> Option<&str> {
+    match op {
+        CanonicalOperation::RecordUpsert { collection_id, .. }
+        | CanonicalOperation::RecordDelete { collection_id, .. } => Some(collection_id.as_str()),
+        _ => None,
+    }
 }
 
 impl FramedTableWalAppender {
@@ -56,6 +75,7 @@ impl FramedTableWalAppender {
             path,
             next_sequence: AtomicU64::new(recovery.last_sequence),
             append_lock: Mutex::new(()),
+            collection_max: parking_lot::Mutex::new(std::collections::HashMap::new()),
         })
     }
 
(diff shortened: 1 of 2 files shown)


vjsingh1984:
## Measured (SF=0.01, DEFAULT config — fix active, no env override)

The fix closes the gap completely in the default config (no `PROXIMADB_OLAP_DELTA_MERGE=0` needed):

| route | pre-fix median | **post-fix median** | DuckDB |
|---|---|---|---|
| tpch | 604 ms | **7 ms** | 10 ms |
| tpcds | 620 ms | **7 ms** | 13 ms |

**86–89× faster, and DF is now FASTER than DuckDB** — matching the `olap_delta_merge=0` baseline (6 ms).

Mechanism confirmed: `open_ms` collapsed from ~1780 ms (Q8) to **0–1 ms** — the per-collection high-water fast-path skips the `changed_oids_since` scan, so wall time now reflects the actual join execution (5–9 ms) instead of the per-table-per-query WAL scan.

`bytes_read` unchanged (~46 MB tpch) — the fix is pure CPU/scan-avoidance, no I/O change. Conformance unchanged (22/22 tpch, 63/63 tpcds).

vjsingh1984:
## SF=0.1 confirms the fix at scale (DEFAULT config)

| route | pre-fix median | **post-fix median** | DuckDB |
|---|---|---|---|
| tpch | 5762 ms | **12 ms** | 15 ms |
| tpcds | 5772 ms | **10 ms** | 16 ms |

**~480–577× faster, and DF is faster than DuckDB at SF=0.1 too.** The gap closes at both scales (SF=0.01: 7 ms; SF=0.1: 10–12 ms) in the default config — no `PROXIMADB_OLAP_DELTA_MERGE=0` needed. Conformance unchanged (22/22 tpch, 63/63 tpcds).

PR #923, opened 2026-07-13
feat(route): TD-ROUTE-1 — capability-aware override candidates (join-bearing plans never flip to Volcano)

vjsingh1984:
## Why

The pgwire override eval (#766) found the hard blocker gating `PROXIMADB_ROUTE_COST_OVERRIDE`: flipping an `olap/parquet` **JOIN** to Native/Volcano errors (Volcano has no join executor). The freshness-safe candidate set was *capability-unaware* — and the TD's suggested fix (exclude by operation class) turns out to be **insufficient**: a `JOIN … GROUP BY` classifies as `op=grp` because `query_operation_class` never inspects the FROM clause, and ADR-058's ancestor-fallback consults can surface Native from a coarse class carrying no operation signal at all.

## What (ADR-058: eligibility precedes selection)

- **`QueryShape::join_bearing`** — dedicated eligibility bit computed at the pgwire boundary (`query_join_bearing`: multi-table FROM / explicit JOINs; **fail-closed** for set-op / non-SELECT bodies), wired at all three `route_select_advised` call sites.
- **Decision-time gate**: all three flip paths — explore, TD-170 RTT-budget hard gate, min-advantage soft gate — refuse Native as a flip **target** for join-bearing plans, covering the ancestor-fallback hole. Static routes untouched; with the override OFF (the default) behavior is identical.
- **Observability**: the eligibility veto is disclosed in the decision reason (`… flip to Native suppressed — join-bearing plan …`) for EXPLAIN/telemetry.
- **Eval is now the enable gate it was designed to be**: the join case flipped from tolerated-error to a **strict parity assert** — verified locally: under the override the join executes on DataFusion and matches the baseline (PASS, 3s).
- 3 new unit tests (flip suppression + disclosure, explore skip, RTT-gate eligibility); **all 63 routing unit tests pass**, zero regressions. Also fixed a stale comment referencing a nonexistent `compute_scheduler::override_candidates`.

## Pre-existing develop conditions (not regressions)

- **TD-ROUTE-2 filed**: the eval **stack-overflows at the default test stack on clean develop** (verified by stash A/B at a0fa74f95; passes in 3s with `RUST_MIN_STACK=16777216`). The enable-gate eval must become runnable at stock settings — bisect + fix tracked there.
- `bench_cross_modal_fusion_e2e` and `bench_14_graph_operations` fail to compile on clean develop (FilesystemFactory / graph API drift) — untouched by this PR.

## Verification

- `cargo nextest run --lib compute_scheduler route_cost_model` — 63/63 pass
- `RUST_MIN_STACK=16777216 cargo nextest run --test route_cost_override_pgwire_eval` — PASS (strict join parity under override)
- `make work-commit-check` green; `cargo clippy --lib --bins -- -D warnings` clean; `cargo fmt --check` clean; `check_td_adr_unique.py` 0 errors

## Enablement path (remaining in TD-ROUTE-1)

Run the eval + TPC-H/TPC-DS ratchets green under override-ON, then decide the default.


vjsingh1984:
Your 4 failing jobs (Rust Compile lib/test, Tests unit/doc — 'Set up job', ~2s) are not your change: `mozilla-actions/sccache-action@v0` from #920 doesn't resolve (no v0 tag exists) and fails every Rust-touching PR at setup. Fix: #925 (pins @v0.0.10). Re-run your failed jobs after it merges.

PR #940, opened 2026-07-13
fix(route): TD-ROUTE-2 eval stack pin + review follow-ups to the #923 capability gate

vjsingh1984:
## What

Two halves, one routing area (fresh-context review of merged #923 + the TD-ROUTE-2 harness fix):

### TD-ROUTE-2: enable-gate eval runs at stock settings
The override eval (TD-ROUTE-1's enable gate) now runs its body on a dedicated 16 MiB thread with a current-thread runtime (same `#[tokio::test]` semantics) — passes in ~3s with no `RUST_MIN_STACK`. Was: stack-overflow abort at the default test stack, pre-existing on clean develop. Root-cause bisect stays open in TD-ROUTE-2.

### Review follow-ups to #923 (fresh-context reviewer; no correctness blocker found, four real refinements)
1. **`query_join_bearing` recurses** through derived tables, nested joins, CTEs, and WHERE/projection subqueries (PATH B lowers these to semi/anti/left joins) — the top-level-FROM walk missed them. Not a result bug (the #853 DataFusion floor serves gap-missed flips correctly), but such a flip **stamped Native into the io_trace while DataFusion executed — poisoning the cost model's per-backend EWMA cells** — and `EXPLAIN ANALYZE` diverged. 12-case unit test added.
2. **Premise corrected everywhere**: Volcano *does* have hash/nested-loop join executors; the real capability gap is **no multi-table external-parquet scan on the Native arm** (the vectorized native-over-parquet path is single-scan-only). The old "no join executor" wording would have misdirected future capability work (e.g. the planned `vectorized_supports`-style eligibility).
3. **Suppression disclosure reworded** to claim ineligibility rather than a would-be flip (the cost gates aren't re-checked in that branch).
4. **Decision-level eval assert (6c)**: with the warmed global model and override ON, a join-bearing `olap/parquet` shape must keep the static DataFusion route with no `OVERRIDE`/`EXPLORE` in the reason — result parity alone passes even with the gate reverted (the floor hides it), so the gate's e2e regression coverage now lives in the eval itself.

Also repairs a doc-comment misattachment from the #923 insert (cardinality-hint + operation-class docs had merged onto `query_join_bearing`) and updates TD-ROUTE-1/TD-ROUTE-2 statuses.

## Verification

- 44/44 routing + pipeline unit tests (`cargo nextest run --lib compute_scheduler relational_pipeline`), incl. the new `join_bearing_detects_joins_at_every_nesting_level`
- Eval passes **at stock settings** (2.6s) including the new decision-level assert
- `make work-commit-check` green; `cargo clippy --lib --bins -- -D warnings` clean; `cargo fmt --check` clean; TD ids unique


vjsingh1984:
**Enable-decision evidence (local, 2026-07-13, this branch):** the pgwire conformance ratchets hold under `PROXIMADB_ROUTE_COST_OVERRIDE=1` with the capability gate active — TPC-H baseline PASS (2.60s), TPC-H override-ON PASS (2.66s), TPC-DS override-ON PASS (2.53s). Caveat for the record: within a single run the model starts cold and warms as queries execute, so this proves *no regression under override-ON*, not that every flip fired — deliberate-flip parity + decision-level coverage lives in `route_cost_override_pgwire_eval` (also green at stock settings). Combined, this is the evidence package TD-ROUTE-1's enable decision calls for; remaining before a default flip: run the same under CI conditions + TD-ROUTE-2's root-cause bisect.

PR #1016, opened 2026-07-15
docs(adr): ADR-062 + TD-RDSTRAT-6 full spec — coalesced RaBitQ-header + self-describing footer PAX layout (REVIEW BEFORE CODE)

vjsingh1984:
## What this is

**Spec-first, review-gated — NO CODE in this PR.** The full architecture decision + implementation spec for TD-RDSTRAT-6 (the coalesced-binary-tier redesign the 2026-07-15 eval motivated). Code (PR1) opens only after this review signs off.

- **ADR-062** (Proposed → Accepted on review): decision + rationale + alternatives + consequences.
- **TD-RDSTRAT-6**: upgraded from the Proposed memo to a full implementation spec (decision rationale defers to the ADR to avoid duplication).

## The decision in one paragraph

The WS8 VOE block-centroid-prune gate is **GET-capped, not recall-capped** (measured in #1015: block-prune can't reach both high recall and low GETs; root cause: PAX is block-major → RaBitQ read per block; the cascade cuts bytes not GETs; `per_get=20 ≫ per_mib=5`). ADR-062 hoists RaBitQ into a **coalesced header region @0** (scan at keep=100% → ~0.99 recall, zero prune loss, cached/file), **reranks survivors** with the unchanged SQ8/fp32/metadata/text block columns (reusing `plan_selected_block_ranges` + `read_ranges`), over the **logical-file unit** (per-object DR/lifecycle/versioning/conditional-write atomicity). A **Parquet-style self-describing footer-index** carries schema + block table + per-stripe encoding map + per-block OID bloom (extensible `StatsKind`). Block geometry **self-derives** from dim × tier-composition × backend IOPS; **file-count = query GET-count knob** via GET-budget compaction (the WLP-7 seam). In-place writer + readability-preserving reader per the ADR-061 pre-GA amendment.

## What I want a reviewer to stress-test (open questions)

1. **Header-first write correctness** — flush buffers the batch; confirm no single-row streaming path bypasses the buffer.
2. **Footer-index size at scale** — OID bloom + block table for thousands of blocks; when to spill blooms to a separate region.
3. **Schema-footer vs catalog authority** — v1 footer = write-time snapshot; any read-path assuming catalog-only schema? (ADR-047/048 interaction)
4. **GET-budget compaction trigger** — file-count threshold vs projected-GETs vs hybrid; write-amplification bounds; the WLP-7 scheduler seam.
5. **Conditional-write availability** — S3 `If-None-Match` (GA 2024) / ABFS for the compaction CAS; fallback where unavailable.
6. **Bloom sizing** — FPP target + bits/elt for OID; per-block vs per-file; cost in the cached footer.
7. **Readability during docs→code window** — the footer presence-field preserves legacy readability (ADR-061 amendment); in-place change is atomic per PR.
8. **Phasing** — PR1 (layout+read-path) / PR2 (cost-control) / PR3 (point-lookup) coherent and independently shippable?

## Alternatives captured in the ADR (rejected)

Block-centroid pruning (GET-capped) · read-path-only coalescing (keeps recall loss) · collection RaBitQ sidecar (breaks per-object atomicity) · RaBitQ in footer (serialises hot scan) · per-encoding magic bumps (footer encoding map instead) · cap-files-at-1 (forces over-compaction).

## Validation
- `check_td_adr_unique.py` → 0 errors (62 ADRs / 223 TDs)
- `check_td_status_consistency.py` → passed
- `cargo fmt --check` → clean
- docs-only; no build impacted

## Measured grounding
link:../blob/develop/docs/_internal/status/SIFT1M_VOE_COMPACTED_RECALL_PARETO_2026_07_15.adoc[SIFT1M eval (#1015)] — the block-size sweep that proved the GET ceiling.


vjsingh1984:
# Review — ADR-062 / TD-RDSTRAT-6 (coalesced RaBitQ-header + self-describing footer PAX layout)

**Verdict: Request changes (spec-accuracy) — the *decision* is sound and Accept-worthy; unblock code PR1 once the corrections below land.** The thesis (block-prune is GET-capped, not recall-capped → hoist RaBitQ to a coalesced head region, scan at keep=100%, rerank survivors via the unchanged block columns) is correct, is grounded in *merged* measured evidence (#1015), and reuses infra that actually exists. None of the issues are design flaws — they are factual-reference fixes plus one load-bearing dependency the spec must name. I did not approve; that's the maintainer's call.

---

## What checks out (load-bearing claims verified against code + the merged eval)

- **Cost model** — `per_get=20.0`, `per_mib_read=5.0` confirmed at `src/query/route_cost_model.rs:489-490`. The "`per_get ≫ per_mib`" premise holds.
- **Reuse APIs all exist, at the cited lines:** `plan_selected_block_ranges` (`sst_query_engine.rs:733`, exact), `ObjectRangeCoalescePolicy` (`sst_query_engine.rs:289`), `FileSystem::read_ranges` (`crates/storage/proximadb-storage-filesystem-types/src/lib.rs:507`, exact), `try_pax_cascade` (`search/mod.rs:333`), `PaxBlockWriter` (`block-format/src/writer.rs:134`), `SegmentMeta` (`block-format/src/stripe.rs:357` — correctly stated as *gaining* `rabitq_off/len`), `pax_write_outcome` (`flush/mod.rs:970`, exact), and the mirror tests `write_pax_segment_round_trips_through_reader` / `detection_defaults_to_legacy_and_guards_false_positives` (`segment_format.rs:856` / `:830`, exact).
- **Evidence table matches the merged doc.** `docs/_internal/status/SIFT1M_VOE_COMPACTED_RECALL_PARETO_2026_07_15.adoc` (merged via #1015, same blob on develop and the branch) confirms every number: pruned recall 0.8087 / cascade 0.9891; default ~28 vec/block → 0.867 @ 220 GETs; 8 vec → 0.991 @ 62 530 GETs; 64 vec → 0.991 @ 7 870 GETs; "no sweet spot." The eval's own §"Result 3" *independently* recommends the exact hoist-RaBitQ solution ("~1 GET binary scan + a handful for survivors, SPANN/DiskANN pattern"). The ADR faithfully implements its evidence.
- **ADR cross-refs are accurate.** ADR-061 *does* carry the pre-GA amendment (lines 17-24: "improved internals REPLACE their predecessors… on-disk **readability** of already-written segments/sidecars [preserved]") — the in-place-writer + presence-field-reader reconciliation is genuinely sanctioned. ADR-052 (bytes/round-trips dominate), ADR-020 (WAL authority; segment is a projection/advisory), ADR-036 (per-object access tier) all say what the ADR claims.
- **RaBitQ sizing math is right:** `dim/8+8` → 24 B/row → 24 MB @1M×128d, 200 MB @1536d; SQ8 128 MB, fp32 512 MB. The "index is small/cacheable, data is large/fetch-on-demand" split holds.

---

## Required corrections (factual — fix before merge)

**1. Hallucinated function `cluster_order_pca_ivf`.** No such symbol exists. The real producer is `cluster_order` (`src/storage/engines/sst/block_cluster.rs:101`), which orders by **sign-Gray-code of (v−mean)**, *not* "IVF-cell ▸ Hilbert/PC1." The file's own header (`block_cluster.rs:8-10`) and the fn doc (`:100`) are explicit: *"streaming, no-k-means space-filling proxy for angular locality… true IVF/k-means clustering is the **S4 upgrade**."* The bogus name appears 3× (ADR D3, TD byte-layout, TD compaction). Replace with `cluster_order`.

**2. Over-claim: the IVF clustering producer is "shipped."** ADR §rationale: *"The clustering producer (`PROXIMADB_PAX_FLUSH_CLUSTER=ivf`, shipped) is the input."* Both halves are wrong: (a) the env var `PROXIMADB_PAX_FLUSH_CLUSTER` does not exist — the flush-clustering gate is `block_cluster_enabled()` (`segment_format.rs:180`); (b) the *shipped* ordering is the sign-Gray proxy, not IVF. The eval's `pca/ivf` Pareto row is a measurement-harness clustering, **not** the shipped flush producer. This matters because **D3's rerank-coalescing GET-win depends on survivor locality**: with only the coarse sign-Gray ordering in place, "survivors fall in a handful of adjacent blocks" is weaker than stated. Please (i) correct the name/flag, (ii) state that **S4 IVF/k-means clustering is a co-requisite** for the full D3 coalescing win (D2's ~1-GET scan is independent of it and stands regardless), and (iii) require PR1 to **measure `range_gets`/query on the shipped clustering**, not assume the IVF-clustered ideal.

**3. `read_ranges` is not a batched multi-range GET by default.** The trait default (`filesystem-types/src/lib.rs:507`) is a **sequential loop of `read_range`** ("Default implementation calls read_range for each range"). The GET *reduction* comes from `ObjectRangeCoalescePolicy` merging adjacent survivor ranges into **fewer, larger** ranges (which again needs clustering) — not from `read_ranges` itself. Reword D3/TD so credit goes to the coalescing policy; don't imply `read_ranges` collapses round-trips on its own.

**4. `storage_profile.iops_target` / `iops_min` / `iops_max` don't exist, and conflate two orthogonal things.** `StorageProfile` (`crates/storage/proximadb-storage-common/src/storage_profile.rs:33`) is a **workload** enum (`AppendBulk`/`Churn`, resolved from tags) with **no IOPS fields**; there is no `iops` anywhere in that crate. D6's `block_bytes_target = storage_profile.iops_target ∈ [iops_min,iops_max]` mixes a per-collection *workload* profile with a per-backend *physical IOPS* descriptor (S3/ABFS/MinIO). Introduce a distinct backend-IOPS descriptor (or read it from the object-store backend) rather than off `StorageProfile`.

**5. `PaxSegmentScanner` crate mis-attribution.** The TD Component field lists it under `crates/storage/proximadb-block-format`; it actually lives in `proximadb_storage_common::pax_block` (`crates/storage/proximadb-storage-common/src/pax_block.rs:887`). Minor, but will misdirect the implementer. (`PaxBlockWriter`/`SegmentMeta` *are* in block-format.)

**6. Conditional-write CAS is net-new on cloud backends.** Only a **create-only** primitive exists — `put_if_absent` (`PutMode::Create`, `crates/storage/proximadb-object-store/src/lib.rs`), whose own doc says *"cloud backends need conditional-put support."* There is **no `If-None-Match` / compare-and-swap update path** (grep-clean). D6's "single-object conditional replace (S3/ABFS `If-None-Match`)" is not implemented for cloud today. PR2 must budget wiring conditional-PUT through `object_store`, and the spec should state the **fallback where unavailable**: write-new-under-fresh-name + atomic pointer/directory swap (the Iceberg-style manifest commit `put_if_absent` already supports), never in-place overwrite.

---

## Per-open-question findings

1. **Header-first write correctness — CONFIRMED SAFE.** The legacy single-row/streaming write path is *retired* (`flush/mod.rs:151-377`, ADR-049 M1-3 — "the legacy ProximaBlocks streaming write is retired… re-encodes as PAX"); PAX flush buffers a whole batch, and there is no append/single-row PAX writer in `segment_format.rs`. Computing the RaBitQ centroid over the buffered batch and emitting header-first is compatible. No streaming path bypasses the buffer.
2. **Footer size at scale — reasonable, but state a cap.** Back-of-envelope at 1M vectors: ~300 blocks @128d / ~1800 @1536d × (rows/block × ~10 bits) ≈ 1–2 MB of blooms. Fine cached at 1M; the concern is real at thousands of blocks / high-dim. Give a concrete spill rule (e.g., footer > N MB → blooms move to a dedicated region indexed by the footer, fetched on point-lookup only).
3. **Schema-footer vs catalog authority — consistent, add one invariant.** ADR-020 makes the segment a projection/advisory, so a write-time footer *snapshot* is fine. But ADR-048 (Proposed) D1 ("no path re-derives schema from a v1 config when a typed catalog schema exists") means the reader must **never prefer the footer schema over the catalog when they diverge** (post-evolution). Add that as an explicit PR1 invariant: footer schema = self-description/external-tools/recovery only; catalog wins on conflict.
4. **GET-budget compaction — coherent; add a WA bound + read-driven trigger.** Hybrid (projected-GETs OR file-count 3–5) and the WLP-7 reframe are sound. Missing: an explicit **write-amplification ceiling** and the tie to ADR-061 — compaction is armed for `AppendBulk` per the arm-defaults amendment, but the *GET-budget* trigger should be **read-driven** so a write-only `AppendBulk` collection that is never queried doesn't pay merge cost. State both.
5. **Conditional-write availability — see Required #6.** Net-new for cloud; needs a named fallback.
6. **Bloom sizing — fine; state the target.** 10 bits/elt ≈ ~1% FPP; the load-bearing property (no false negatives → safe skip-to-block) holds for any bloom. Per-block is correct. Please state the FPP target explicitly (1% ⇒ ~9.6 bits/elt) and the per-file cost formula so Q2's cap is derivable.
7. **Readability during docs→code window — CONFIRMED sound.** ADR-061 (lines 17-24) sanctions in-place writer + readability-preserving reader; the presence-field discriminator (footer/header `rabitq_off` present → scan-then-rerank; absent → legacy in-block) is a valid mixed-read marker. Same-magic/version is safe *because the reader branches on presence, not version*. One caveat: route the presence check through the existing `detection_defaults_to_legacy_and_guards_false_positives` guard (`segment_format.rs:830`) so a legacy segment's bytes can't false-positive as a `rabitq_off`.
8. **Phasing — coherent and independently shippable.** PR1 (layout + scan-then-rerank + self-derived geometry + in-place writer/reader) is the indivisible core and delivers the headline GET win on its own (D2 is clustering-independent). PR2 (compaction + cache) and PR3 (OID bloom point-lookup) are cleanly separable. **Caveat (ties to Required #2):** PR1's *D3* rerank-coalescing win is contingent on clustering quality — PR1's acceptance ratchet must measure actual `range_gets`/query on the shipped sign-Gray ordering and report it, not assume the IVF ideal.

---

## Alternatives — correctly characterized

Block-prune (GET-capped ~0.87, matches the eval), collection sidecar (breaks per-object DR/lifecycle/versioning/conditional-write atomicity), RaBitQ-in-footer (serializes the hot scan on a metadata GET), read-path-only coalescing (keeps keep<100% recall loss — correctly flagged as a possible fast-follow, not the end-state), per-encoding magic vs footer encoding-map, and cap-at-1 are all genuinely inferior and fairly stated. No strawmen.

---

## Bottom line

Architecture: **Accept-worthy** — dissolves a *measured* ceiling by moving the *dominant* cost term, reuses real infra, and is mixed-read-safe per a real ADR-061 amendment. Merge-blockers are the factual fixes #1 (hallucinated fn), #2 (clustering over-claim + name the S4 co-requisite), #4 (non-existent `iops_*` fields), and #6 (CAS is net-new — name the fallback); #3/#5 are wording. With those corrected, **code PR1 may proceed** — and PR1 should measure GETs on the shipped clustering so the D3 claim is evidence-gated, not assumed.

PR #1052, opened 2026-07-16
perf(sst): adaptive survivor pool M + tighter coalescing (ADR-062 PR2 p1/p2)

vjsingh1984:
Two co-designed fixes for the coalesced RaBitQ layout's recall-at-scale problem (measured: recall 0.991@100k → 0.968@1M with fixed M=1000).

## Changes

**1. Adaptive M** — the survivor pool scales with the segment's row count N:
- `pax_rabitq_pool_for_top_k(k, n)` = `max(k·mult, ceil(n·rate), min)` with `PROXIMADB_PAX_RABITQ_POOL_RATE` (default 0.01 = 1% survivor fraction).
- The pool decision moved INTO `rabitq_search_segment_coalesced` using `region.n_rows()` (already parsed) — zero extra reads.
- At 1M: M scales from the fixed 1000 → 10000 (maintaining the ~1% fraction that gave 0.991@100k).

**2. Tighter coalescing** — `ObjectRangeCoalescePolicy` for the coalesced read path: gap 64KB→256KB, range 8MB→16MB (env-overridable `PROXIMADB_PAX_COALESCE_GAP`/`_RANGE`). Bytes are ~free on same-AZ object storage → merge survivors more aggressively to keep GET count down despite the larger M.

Also opportunistically fixes a pre-existing develop test breakage (`max_row_group_size()` → `max_row_group_row_count()` + `Option` wrap in `schema_builder.rs`).

## Verification
- 100k regression: **recall 0.991, 19 GETs/query — unchanged** (M=1000 via the `min` floor at 100k).
- fmt / clippy `--lib --bins` (-D warnings) / `make work-commit-check` clean.
- **The 1M gate** (does M=10000 lift recall to ≥0.98?) needs the full SIFT1M base (~516MB; the local dataset is only the 100k subset). The 1M measurement is the qa-gate CI job or a follow-up once the full dataset is available.

## Plan reference
This is PR2 items 2+3 from the approved plan (`/Users/vijaysingh/.claude/plans/toasty-dreaming-pretzel.md`). PR2-1 (per-segment invariants cache — the GET elision win) is next; PR2-4/5 (compaction/per-cell) is gated on the 1M measurement.

### DIFF
--- a/docs/_internal/roadmap/BENCHMARK_EVIDENCE.toml
+++ b/docs/_internal/roadmap/BENCHMARK_EVIDENCE.toml
@@ -61,6 +61,23 @@ version = "develop @ 3a703d0d0 (+ ADR-062/TD-RDSTRAT-6 PR1: coalesced_rabitq fea
 notes = "Single-file collection (one flush batch) with the IVF flush opt-in ON (PROXIMADB_PAX_FLUSH_CLUSTER=ivf) — isolates the per-file ~1 RaBitQ GET win from the PR2 GET-budget compaction. N=100000 subset, 100 queries, top-10, Euclidean, brute-force ground truth. Measured recall@10 = 0.984, range_gets/query = 20. Indicative dev-machine run, not an SLA; bytes_read/query (31.7 MB) is the deferred post-PR1 lever (clustered SQ8/fp32 compression)."
 
 [[claims]]
+id = "coalesced_rabitq_adaptive_m_sift1m"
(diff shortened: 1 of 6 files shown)


vjsingh1984:
Superseded by #1060 — the rebased branch contains all of #1052's adaptive-M + cache changes + the block-geometry/IopsBudget/zstd work, directly on latest develop (picked up the proximadb::embedded fix from #1055).

PR #1069, opened 2026-07-17
feat(sst): ADR-065 Region B (SQ8 hoist) + cache-co-design stack

vjsingh1984:
Hoists SQ8 rerank codes out of PAX blocks into a coalesced same-file Region B so survivor rerank fetches read pure dense SQ8 (not full blocks with OID/props/fp32). Plus the cache-co-design sub-PR (footer Sq8Params mirror, IOP-aligned max_range, CacheTier enum + per-tier trace, byte-budgeted cache) + prefix redundancy fixes.

Layout: [header 56B][Region A: RaBitQ][Region B: SQ8][Region D: row blocks][footer]. Blocks lose the EMBED_BASE SQ8 stripe (hoist_vector_tier flag, keeps f32 for PR4). Two-phase read: rank Region A -> rerank survivors via dense SQ8 ranges -> fetch top-k OIDs from Region D.

IVF k=blockcount (cells~=IOP, ~30 SIFT1M). Morton-on-SQ8 = cold-start bootstrap. No version bumps (pre-release, modify in place).

Measured (SIFT1M, Azure IOP 4MB): recall 0.989, ~18 GETs/q (FS-trace ground truth), 88 MB. Note: io-trace snap.range_gets double-counts 2x (measurement bug, follow-up).

Follow-ups: io-trace double-count fix, NVMe survivor-range cache (ADR-065 Q3), IVF cheapening (SIMD + mini-batch + train-at-compaction TD-WLP-4b).

### DIFF
--- a/crates/foundation/proximadb-clustering-kernel/src/lib.rs
+++ b/crates/foundation/proximadb-clustering-kernel/src/lib.rs
@@ -10,10 +10,31 @@
 use anyhow::Result;
 use rand::{Rng, SeedableRng, rngs::StdRng};
 
-/// Inline squared L2 distance (avoids distance-engine overhead in the
-/// clustering inner loops — this is a write-time, low-cardinality path).
+/// Squared L2 distance — the clustering inner-loop hot path (k-means++ init,
+/// Lloyd assignment, convergence check, final assign). On aarch64 this
+/// dispatches to a hand-NEON kernel (TD-WLP-4b: ~3.8x faster than the
+/// auto-vectorized scalar loop at dim=128 on an M1 Max — the iterator-chain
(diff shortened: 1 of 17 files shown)


vjsingh1984:
Heads-up from the CI sweep: this branch is now **CONFLICTING with develop** (4-file overlap after #1064 TD-RDSTRAT-7 merged: `block-format/writer.rs`, `pax_block.rs`, `segment_layout.rs`, `sst/segment_format.rs` — both PRs rewrite the PAX write path). A trial rebase confirms the conflicts are semantic (Region B hoist vs lossless clustered compression), so they need the driving session's resolution rather than a mechanical fix. Auto-merge is armed but can't fire while DIRTY; after rebasing, CI will also revalidate against the post-#1064 writer, which the previously-green run never compiled against.

PR #1078, opened 2026-07-17
docs(td): TD-RDSTRAT-8 — two-level IVF (coarse-probe) on NOVA

vjsingh1984:
## TD-RDSTRAT-8 — Two-level IVF (coarse-probe) on NOVA

Measurement-motivated research TD (no code). Single-level IVF uses `cells = blockcount`
(`k ∝ N`), so GETs/query scale **O(N)** with collection size — fine at SIFT1M (~50 GETs),
broken at 10M+ (~500 projected). The SIFT sweeps in this TD's *Why* rule out the two
adjacent levers as GET reducers:

- **Survivor pool M** — GETs move only 8→10 across a 100× M range; M is bytes/recall, not GET.
- **k tuning** — GETs near-flat 50–54 across k 8–128 (U-shape, min k=64).

Two-level IVF (coarse quantizer `k_c≈√N`, probe `nprobe` fixed) drops the exponent
**O(N)→O(nprobe)≈O(1)** (FAISS model). **Target: NOVA** (production-ready, hierarchical /
progressive substrate; SWIFT deprecated). SST keeps single-level IVF so users choose by
scale. Open questions + next action in the file.

### DIFF
--- a/docs/10-quality/td/TD-RDSTRAT-8-two-level-ivf-coarse-probe.adoc
+++ b/docs/10-quality/td/TD-RDSTRAT-8-two-level-ivf-coarse-probe.adoc
@@ -1,0 +1,94 @@
+// Copyright (C) 2026 ProximaDB
+// SPDX-License-Identifier: Apache-2.0
+= TD-RDSTRAT-8: Two-level IVF (coarse-probe) on NOVA — break the GETs∝N scaling of single-level IVF
+
+[cols="1,3", options="header"]
+|===
+| Field | Value
+| Status | Open (research/proposed — not yet implemented; measurement-motivated).
+| Priority | P2 — unblocks ANN GET-cost at 10M+ scale; not blocking at SIFT1M (single-level IVF is ~50 GETs/q there).
+| Severity | n/a (new capability, no regression).
+| Component | `src/storage/engines/nova/` (progressive search / hierarchical pruning — the coarse-probe host); SST keeps single-level IVF (`src/storage/engines/sst/segment_format.rs::rabitq_search_segment_coalesced`).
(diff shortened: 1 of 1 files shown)


vjsingh1984:
**Rev 2 pushed** — critique + full implementation spec after a code review of NOVA vs SST:

- **Host redirected NOVA → SST.** NOVA's substrate claims don't hold on `develop`: SuperBlocks are a stub, progressive search is full-scan staged *filtering* (no cell routing), zone maps aren't hot-path pruning, no PAX/region/coalescing machinery (plain Parquet row-groups), and engine choice is immutable per collection — a re-ingest cliff at exactly the 5–10M boundary this TD exists to serve. SST already has ~90% of the substrate.
- **Design: v3 segment layout composing with ADR-065** — new Region A0 coarse directory (persisted PCA model + PCA-space centroids + radii + per-cell region offsets, ~120–300 KB cached), regions reordered (coarse, fine, row), block padding at cell boundaries. This fixes **both** O(N) terms — survivor GETs *and* the Region-A whole-region fetch (rev 1 missed the second: at 10M the fetch-all scan is ~160–240 MB/query even with two-level survivors).
- **Recall guard**: bound-based nprobe escalation (LB = d_pca − radius vs d_k·(1+ε)) replaces keep=100%; the eval gate stays the recall authority (TD-WLP-9 pattern). Production flush stays sign-bit L0 per ADR-065; v3 is compaction-written only → zero migration events, mixed-read safe, default OFF.
- **Execution plan**: 4 PRs (format+writer / reader / escalation+cache / scale+hardening) with per-PR acceptance criteria, env knobs, io_trace additions, and eval gates incl. the 10M ADR-0044 downstream unlock.

PR #1082, opened 2026-07-17
fix(objstore): staged segment I/O — kill the URL-string-strip class in flush/compaction/reads (TDD)

vjsingh1984:
## What

Comprehensively closes the URL-as-local-path (string-strip) defect class in SST segment I/O, TDD-first. Found via the #1080 QA-run artifact: the **normal** flush on cloud bases still wrote segments to a literal local `adls:/` path (#1061 fixed recovery staging only), and the sweep found the same class in **compaction** (armed-by-default; false-success would delete the input segments) and **four segment-read sites**.

- **RED first**: `graceful_flush_persists_segment_to_object_store` (Azurite process test) — WAL+manifest in the store, segment absent, artifact dir present.
- **One sanctioned boundary**: `sst/staged_write.rs` — `StagedSegmentWrite` (write: local scratch → upload on finalize; zero-copy for local bases incl. #1061's recovery tempdir), `read_object_bytes` + `LocalizedSegment` (read: FileSystem-routed; scratch removed on Drop).
- Applied to both flush arms, the compaction output write, and all four search read sites.
- **Source-guard**: zero `strip_prefix("file://")` in flush/compaction/search — the class cannot silently return.

## Verified (Azurite, identical local/CI script path)

New test RED→GREEN · full restart suite 3/3 (3-boot, zero-vector batch-3 gate, graceful-flush) · clean re-run leaves **no** `adls:/` artifact · primitive units + source-guard green · clippy `-D warnings` + fmt clean.

## Notes

Composes with #1065 (gates) and #1080 (QA/nightly tiers). `nova`/`swift` extraction + DataFusion `pax_adapter` file:// strips are genuinely-local input paths today — flagged in the TD for audit, not segment-I/O load-bearing.

### DIFF
--- a/apps/proximadb-server/tests/object_store_restart_recovery.rs
+++ b/apps/proximadb-server/tests/object_store_restart_recovery.rs
@@ -475,15 +475,6 @@ async fn cold_recall_ratchet_survives_restart() -> anyhow::Result<()> {
     const PROBES: usize = 8;
     const MIN_RECALL: f32 = 0.9;
 
-/// TD-OBJSTORE-4 defect-6 redux (task: normal-flush string-strip): a GRACEFUL
-/// flush on a cloud base must persist the segment INTO the object store — and
-/// must NOT write it to a literal local `adls:...` directory (the
-/// URL-as-local-path artifact). #1061 fixed the RECOVERY staging path only;
-/// the normal flush retained the false-success class (masked by WAL replay).
-/// RED on that state: no segment blob in the store + artifact dir on disk.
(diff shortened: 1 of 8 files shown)


vjsingh1984:
Post-review verification on the rebased branch: full 4-test Azurite battery green (3-boot proof, zero-vector batch-3 gate, cold recall ratchet, graceful-flush persistence — `4 passed`), no `adls:/` artifact after a clean run, clippy `-D warnings` + fmt clean. All review findings (B1 finalize-returns-bytes, S1 Arrow sidecar pair, S2 Drop guards, TD notes) applied. Auto-merge armed per owner.

PR #1175, opened 2026-07-24
fix(sst): compaction passed records.len() as embedding_count — 105k phantom embedding columns (+40.5 GB, 42 s/block) — derive real count + MAX_BLOCK_ROWS bound

vjsingh1984:
## Summary

Fixes the TD-COMPACT-2 memory blow-up root cause, pinned by #1173's `[COMPACT mem]` stage profile (+40,475 MB between `pre_write` and `write_done` at 105k records).

**Root cause:** all three PAX compaction call sites passed `records.len()` as the writer's `embedding_count` — the embedding-*modality* count, normally 1. `PaxBlockWriter` therefore allocated one embedding-column buffer per "modality" = **105,000 columns**, pushing 105k `Option<Vec<f32>>` (24 B each) per record (~2.5 MB/record transient) and serializing 105k all-null stripes at every block cut (~42 s/block). The flush path passes the correct count and was always healthy — which is why only compaction blew up.

## Changes

- `compaction.rs:1391,1535` + `compactor_impl.rs:990`: `embedding_count = max(records[i].embeddings.len()).max(1)`. Footer `embed_count` is informational — not a format change.
- `engine_constants.rs`: `MAX_BLOCK_ROWS = 16*1024`; `pax_block.rs` cuts at the row bound OR the existing byte threshold (defense-in-depth vs degenerate per-row estimates / giant cells). Cell boundaries still force-cut — no block spans two cells; the A0 directory stores block *ranges* and probe/rerank readers are row/byte-granular, so multi-block cells are already format-compatible (no read-path change).
- New test `two_level_row_bound_splits_giant_cells`: 17k-row cell splits into bound-respecting, cell-aligned blocks; A0 extents exact; full round-trip of ids + Region-B vectors.
- Fixed pre-existing stale test `for_path_resolves_by_scheme` (old `CLOUD` preset vs #1060 per-provider presets).
- TD-COMPACT-2 updated with the root-cause verdict.

## Verification

- Suites: block-format 67/67; storage-common 444/448 — the 4 failures are pre-existing (reproduced failing via stash on the unmodified baseline; their post-#1060/#1069 semantics are a separate decision).
- fmt / clippy -D warnings / root `cargo check --lib` clean.
- Recall invariant: block cuts change neither row order nor cell membership; default geometry emits identical cuts — only the write-side transient shrinks (~105,000× fewer embedding buffers).
- Measured before/after on the SIFT 120k compaction sweep: **before** peak RSS 44.8-46.6 GB, encode 17-20 min, ~42 s/block; **after** numbers land as a PR comment when the in-flight verification sweep completes (build was still linking at push time).


vjsingh1984:
**Measured after-numbers (same SIFT 120k compaction sweep, same machine, fixed binary):**

| metric | before (develop @ instrumentation) | after (this PR) |
|---|---|---|
| compaction wall (5 L0 → 1) | 1,018–1,194 s | **2.15 s** (~500×) |
| peak RSS delta during compaction | +40,475 MB | **+374 MB** (~108×) |
| per-block cut/compress | ~42 s | ms-scale |
| outcome at 1M | jetsam kill (~126 GB) | n/a — write-side transient now O(real embedding count) |

`[COMPACT mem]` stage bracket and `[PAX write detail]` buckets on the fixed binary show the write phase back to flush-path-healthy profile (raw_buffer/block in the tens of ms). Verification sweep: `RESULT N=120000 outcome=completed elapsed_s=2 peak_rss_kb=4543408 base_rss_kb=4169056 L0_before=5 files_after=1`.

vjsingh1984:
## Independent 1M verification — fix confirmed

Ran the TD-COMPACT-1 1M SIFT ratchet on this branch merged over develop@bcbcde710 (same box, same paced-ingest protocol as the baseline run earlier tonight):

| Metric | develop baseline (tonight) | this PR |
|---|---|---|
| Peak server RSS | 40–52 GB spikes, one OOM kill | **7.4 GB max** |
| Compactions | 939 s / 1283 s / 538 s (encode-dominated under the phantom-column storm) | **9 compactions, 19.7 s total — mean 2.2 s** |
| 1M ingest wall-clock | 129 s (needed a forced 30 s drain-pause at 40 GB) | **88 s, no pauses** |
| Recall across compaction | 0.999→0.986 dip (original TD-COMPACT-1) | **identical pre/post**: setA 0.9850==0.9850, setB 0.9815==0.9815 (disjoint 200-query sets) |
| Compaction I/O (new S2 counters) | "0.0 MB/s" | 261 MB read / 261 MB written across 9 events |

**One observation to check**: absolute recall is 0.985 vs the pre-storm baseline's 0.999 (measured pre==post, so NOT compaction-caused). A candidate is `MAX_BLOCK_ROWS = 16*1024` now re-cutting the ~21k-row flush cells (the PR's recall-invariance note assumes default geometry emits identical cuts, but 21k-row cells exceed the bound). Could also be plain layout variance across 10-vs-6 segments — worth one targeted look before merge. Above the 0.90 gate either way.

TD-COMPACT-1 is updated to Resolved with this evidence on branch `fix/td-compact1-finalize` (S2 throughput-metric fix + compaction prometheus counters), which will rebase once this lands.

vjsingh1984:
Second commit added (same campaign, one CI cycle): **TD-NET-1 S1+S2** — the internal mux upstream port is now resolved as `[api] internal_mux_port` > `PROXIMADB_INTERNAL_MUX_PORT` > derived `unified_port + 10001` (default deployments still get exactly 15679 — bit-identical backward compat; a second instance derives a unique port automatically), and the internal bind is now **fail-closed**: startup aborts with a descriptive error instead of silently forwarding gRPC/Flight to a foreign instance. Unit + config tests included; root check/fmt/clippy clean.

PR #1234, opened 2026-07-26
feat(search): enable PAX coarse-probe (IVF) by default + TOML threading + geometric nprobe

vjsingh1984:
## What
Enables the PAX coarse-probe (IVF) system **by default** + makes it **TOML-controllable** + wires the **geometric nprobe formula** (`sqrt(ncells) × multiplier`). The COGS measurement arc proved IVF is the master cost lever: full-scan (~100 GETs/q, $105/tenant/mo at 10 q/s) → IVF nprobe≈sqrt(ncells) (~24 GETs/q, ~$25, **~4× COGS cut**, recall ~0.98).

## Why IVF wasn't firing
`PROXIMADB_PAX_WRITE_A0_TRAIN` was **default-OFF** → compaction emitted untrained v1 segments (no A0 coarse directory) → the read-probe (already default-ON) found nothing → full-scan fallback. The nprobe default was also ad-hoc (`k_c/4`), not geometric, and nothing was TOML-controllable.

## Change (6 files)
1. **`CoarseProbeConfig`** sub-struct on `SstConfig` (`[storage.sst_config.coarse_probe]`): `enable_read_probe` (true), `enable_write_train` (**true — the FLIP**), `nprobe_multiplier` (1.0), `nprobe_min` (3), `nprobe_max` (0).
2. **`CoarseProbeSettings` + `OnceLock`** (`segment_format.rs`), init at `SstEngine` boot from config (mirrors `SHARED_INVARIANTS_CACHE`).
3. **3 free functions wired**: env override (call-time) → else TOML settings:
   - `coarse_probe_enabled` → kill-switch → `settings.enable_read_probe`
   - `coarse_probe_nprobe` → explicit → **geometric** `ceil(sqrt(k_c) × multiplier)`
   - `ivf_probe_enabled` → truthy → `settings.enable_write_train`
4. **`ENV_GATE_REGISTRY`** rows 172/173/179 → **config-mirror** tier.
5. `config.toml` documents `[storage.sst_config.coarse_probe]`.

## Properties
- **Mixed-read-safe**: v1 (A0-less) + v3 (A0-trained) segments coexist; the read path probes only when `a0_len > 0`.
- **Geometric nprobe**: `ceil(sqrt(ncells) × 1.0)` replaces `k_c/4` — sub-linear (V^0.25), ~56× fewer ranks at 1M, recall ~0.98. Bump `nprobe_multiplier` (TOML) for higher recall.
- **Precedence**: env (call-time) > TOML `[storage.sst_config.coarse_probe]` > `CoarseProbeConfig::default()`.
- The recall/GET eval gates passed (`nprobe_sweep_trained_1m` ledger claim).
- `cargo check --lib` ✅ (6m) · `work-commit-check` ✅ (env-gates 218).

## Out of scope
- Per-collection `nprobe_sweep.py` refinement (geometric is the safe default; sweep optimizes per dataset).
- Per-tier `nprobe_multiplier` ladder (economy/pro/enterprise).
- The Azurite bench verification (IVF fires + recall ≥ ratchet + GETs drop) — to be run once this lands on develop.


vjsingh1984 (commented):
## Review — strong rate-card win, but a release-blocking ordering dependency

**Rate card: yes, this is the change that makes the undercut real.** It flips `enable_write_train` default-ON, so training compaction (TD-COMPACT-5, #1207) actually fires by default → segments train → cold queries take the probe path **by default** instead of only when an operator sets an env var no customer would set. Measured effect on the default customer path: cold ~180 GETs/query ($7.2/M) → ~83 GETs/query ($3.3/M), a ~2× default cold-COGS cut that directly funds a lower price. Recall is safe — the **SIFT recall ratchet (N=10k) passes** with the new geometric nprobe. Design is clean (TOML→OnceLock mirroring SHARED_INVARIANTS_CACHE, env-overrides-win, and the max-then-min `geometric_nprobe` correctly avoids the clamp panic).

**🔴 BLOCKER — must land #1233 first (or cherry-pick its 2-file fix).** Making write-train default-ON means every trained-collection search fires the coarse probe, which hits a **duplicate IVF metric registration panic** discovered today (TD-METRICS-3, PR #1233): `proximadb_ivf_cells_total` (+3 siblings) are registered in BOTH `sst/metrics.rs` (unlabeled IntCounter) and `consumption_metrics.rs` (per-tenant CounterVec), and the register fallback reused an empty help → the tokio worker panics on the first probe-armed search with the billing observer active. **CI here is green only because unit/integration tests use small data that never trains A0 → never probes → never hits it; production (large data → compaction → probe) crashes on every search.** This PR touches none of the metrics files, so on current develop it ships a crash-on-default-search. Verified: my diverse-query bed panicked pre-fix, then ran 1,200 probe-armed searches clean after #1233.

**Two follow-up notes (not blockers here):**
1. Geometric nprobe recall is ratchet-verified only at N=10k — worth a 1M confirmation before GA (my #1210 sweep used fixed nprobe; the geometric `sqrt(k_c)` gives a different value per k_c).
2. The cold-COGS claims in this PR are correct. Separately, the rate card's **hot** number needs the TD-CACHE-8 correction (PR #1233): 'hot ≈ 0 GET' was a result-cache artifact — real diverse-query hot is ~18 GETs/query (~$0.7/M), not free. Doesn't affect this PR's cold claims.

PR #1237, opened 2026-07-26
docs(adr): ADR-076 async compaction + admission control + training debounce (TD-COMPACT-6/7/8)

vjsingh1984:
## What
Files the **compaction co-design** for cross-model review before implementation. ADR-076 (Proposed) + 3 TDs (Open). **Docs-only — no code changes.**

## The problem (measured 2026-07-26)
ProximaDB's SST compaction is **inline synchronous** — every flush blocks for the full compaction cycle (read L0 → merge → train PCA/IVF → write → delete). During 1M SIFT ingest on Azurite, the TD-COMPACT-5 training arm fired **every ~35s** (5-7 redundant cycles), each costing N object-store GETs + ~35s PCA CPU + 1 PUT + N DELETEs. The result: ingest throughput ~1 flush/35s (should be ~1/s), query latency ~400ms during ingest (should be ~1ms idle), and ~7× wasted object-store I/O.

## The design (ADR-076 — first principles + cross-system evidence)

Every production LSM system uses **async compaction + admission watermarks**. ProximaDB is the outlier:

| | RocksDB | HBase | Cassandra | Delta/Iceberg | **ProximaDB** |
|---|---|---|---|---|---|
| Compaction | Async | Async | Async | Batch maintenance | **Inline (blocks flush)** |
| Admission control | slowdown/stop triggers | throttling + pressure | IO throttle | N/A | **None** |
| Write blocking | No | No | No | No | **Yes** |

**Three decisions (each a boundary, not a micro-optimization):**
- **D1 — Async compaction** (TD-COMPACT-6): move compaction from inline-synchronous to the **existing idle background workers** (2 daemon threads at `engine.rs:212`, currently unused). Flush creates L0 and returns immediately (~1s); compaction runs concurrently. Ingest throughput ~30×.
- **D2 — L0 admission watermarks** (TD-COMPACT-7): RocksDB-style `l0_slowdown_trigger` (delay writes ~1ms) + `l0_stop_trigger` (stall writes) when L0 backlog exceeds thresholds. Prevents compaction storms.
- **D3 — Training-arm debounce** (TD-COMPACT-8): don't fire TD-COMPACT-5 on a segment that received new writes in the last 120s. Eliminates the 7× redundant re-training loop. One post-ingest pass instead of N during-ingest cycles.

**Object-store co-design**: each compaction cycle = N GETs + 1 PUT + N DELETEs. At 50ms/GET (cloud), 7 redundant cycles during ingest = pure waste. Async + debounce → 1 cycle → 7× less compaction I/O.

## Boundary contract (each surface well-integrated)
- **Flush**: memtable → L0 `.pax` (atomic, non-blocking, ~1s). Never awaits compaction.
- **Compaction worker**: reads L0 → merge + train → write L(n+1) → delete L0. Background, concurrent with flushes.
- **Admission control**: monitors L0 count; throttles/stalls writes when backlog exceeds watermarks.
- **Training debounce**: guards the training arm — waits for segment stability (120s no flush).

## Verification (guards)
`check_td_adr_unique` ✅ (76 ADRs, 288 TDs, 0 err) · `check_design_status_index` ✅.

## Requested review focus
1. Is **async compaction** (D1) the right first step vs the Iceberg/Delta scheduled-batch model? (The async-background model reuses existing infrastructure; the scheduled-batch is a larger refactor.)
2. Are the **watermark thresholds** (slowdown=8, stop=12) reasonable defaults? (RocksDB uses 20/36 for L0; we have fewer files per level.)
3. Is the **120s debounce** the right stability window? (Should it be configurable per workload?)
4. Does the **boundary contract** correctly separate flush / compaction / admission / training without overlap or gaps?

### DIFF
--- a/docs/10-quality/td/TD-COMPACT-6-async-compaction.adoc
+++ b/docs/10-quality/td/TD-COMPACT-6-async-compaction.adoc
@@ -1,0 +1,50 @@
+// Copyright (C) 2026 ProximaDB
+// SPDX-License-Identifier: Apache-2.0
+= TD-COMPACT-6: Move compaction from inline-synchronous to async-background
+:status: Open
+:dim: D1
+:pillar: REL
+
+[cols="1,3", options="header"]
+|===
+| Field | Value
+| Status | **Open.** Design decided in ADR-076 (Proposed). First slice — the highest-impact change (ingest throughput ~30×).
+| Priority | **P0** — the inline-blocking pathology gates ingest throughput on object storage (measured: 1 flush/35 s during training compaction).
+| Component | `src/storage/engines/sst/flush/mod.rs:747-781` (inline call); `src/storage/engines/sst/compaction.rs:589-902` (`schedule_compaction` + `worker_loop`, currently idle); `src/storage/engine.rs:208-213` (2 workers, hardcoded).
+| Evidence | 2026-07-26 verification: TD-COMPACT-5 training arm fires every ~35 s during ingest (5-7 cycles). Each cycle: N object-store GETs + PCA (~35 s CPU) + 1 PUT + N DELETEs. Flush blocks for the full duration.
+|===
+
+== Mechanism
+
+The production flush path calls `run_due_compaction` INLINE (`flush/mod.rs:756`,
+awaited). The existing `schedule_compaction` (`compaction.rs:589-628`) +
+`task_queue` + `worker_loop` (`compaction.rs:792-902`) + 2 daemon workers
+(`engine.rs:212`) provide the async machinery — built, tested, and idle.
+
+== Fix
+
+1. Replace the inline `run_due_compaction` call at `flush/mod.rs:747-781` with
+   `schedule_compaction` (enqueue the task; the background worker picks it up).
+2. The flush returns immediately after creating the L0 file (~1 s), without
+   awaiting compaction.
+3. The background worker runs `perform_compaction_enhanced` concurrently with
+   flushes. The atomic coordinator serializes the publish; the
+   `schedule_compaction` per-output dedup (`compaction.rs:596-610`) prevents
+   double-compaction.
+4. The inline path remains as a `#[cfg(test)]` synchronous option (tests can
+   await compaction deterministically).
+
+== Verification
+
+* Ingest throughput: ~1 flush/s (vs ~1 flush/35 s) during 1M SIFT ingest on
+  Azurite.
+* No compaction contention on the query path (idle-server queries ~1 ms during
+  ingest, not ~400 ms).
+* Compaction completes post-ingest (poll: L0 count drops, L1 appears).
+
+== Related
+
+link:ADR-076-async-compaction-admission-control.adoc[ADR-076],
+link:TD-COMPACT-5-training-compaction-trigger.adoc[TD-COMPACT-5],
+link:TD-COMPACT-7-l0-admission-control-watermarks.adoc[TD-COMPACT-7],
+link:TD-COMPACT-8-training-arm-debounce.adoc[TD-COMPACT-8].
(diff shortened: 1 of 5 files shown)


vjsingh1984 (commented):
## Review — direction is right; D1 has a load-bearing prerequisite the ADR misses, and Q4's premise is inverted

Strong problem statement (the 7× redundant training I/O is a real regression in the TD-COMPACT-5 arm I wrote — D3 is a legitimate fix to my design), and the cross-system table is accurate: every production LSM uses async + watermarks. But two things must be corrected before implementation.

### 🔴 D1 blocker: there are TWO `Compaction` instances, and the flush path holds the WORKER-LESS one

The ADR says 'move `run_due_compaction` to the existing background workers … idle because the inline path bypasses them.' That framing is wrong, and the difference is catastrophic if implemented naively:

- `SstEngine.compaction_manager` (**core.rs:433**) is built as `Some(Arc::new(Compaction::new(config)))` with **no `start_workers` call**. The flush inline path (`flush/mod.rs:757`) uses THIS instance via `self.compaction_manager()`.
- `StorageEngine.compaction_manager` (**engine.rs:148 → `start_workers(2)` at :212 → :213**) is a **separate `Compaction` instance** — that's the one with workers.

So the workers don't drain the flush path's manager — they drain a *different manager the flush path never touches*. If D1 just swaps `run_due_compaction` → `self.compaction_manager().schedule_compaction(task)`, tasks land in `SstEngine`'s manager's `task_queue` which has **zero workers** → compaction never runs → L0 grows unbounded, training never fires, recall collapses. **D1's real first step is unifying the two managers** (start workers on the `SstEngine` manager, or route flush scheduling to the worker-bearing one). Please make that explicit as D1's precondition; it's not a reuse, it's a wiring fix.

### 🟡 Q4 is inverted: a per-collection guard already exists, and going async ADDS it

There already IS a per-collection concurrency guard: `active_compactions: RwLock<HashMap<String, SstCompactionTask>>` (compaction.rs:227), consulted by `schedule_compaction` (compaction.rs:602). The **inline** `run_due_compaction` path is the one that does NOT consult it — so today's inline path is the racier one. Going async *through `schedule_compaction`* therefore **improves** the race posture (adds the de-dup), not worsens it. Reframe Q4: verify the async path's `active_compactions` check is keyed/held correctly across the full read→train→swap→delete window (not just at enqueue), and that a flush landing a new L0 mid-compaction is handled (the next `schedule_compaction` should coalesce, not stack).

### Panel answers

1. **Async (D1) over scheduled-batch (Iceberg/Delta)** — async is right for ProximaDB's flush-driven LSM (Iceberg's batch model assumes append-only, no flush→L0 pressure). Reusing the built machinery is correct *once the two-manager wiring is fixed*. Scheduled-batch is a later option, not the first step.
2. **Watermarks 8/12** — reasonable given `max_files_per_level=10` (RocksDB's 20/36 assumes more files/level). But `stop=12 > max_files_per_level=10` is inconsistent — stop should sit at or just below the level cap, else leveling triggers before the stop watermark. Suggest slowdown=6, stop=10 (=cap), and make both `SstConfig` fields (they'll need tuning per backend latency).
3. **120s debounce** — right shape, but make it **adaptive**: debounce = `max(120s, 3× observed mean inter-flush interval)`. A fixed 120s over-waits on slow trickle ingest and under-waits on bursty 1M loads. Fall back to 120s when no flush-rate history exists.
4. Covered above.
5. **Inline path as `#[cfg(test)]`** — keep it, but the *default* tests should exercise the **async** path with a deterministic drain barrier (a test-only `await_quiescent()` that blocks until `task_queue` empties + `active_compactions` clears). Testing only the inline path would leave the production async path — including the two-manager wiring and the watermark feedback loop — uncovered, which is exactly the class of gap that let the metrics-panic and warm-manifest bugs ship.
6. **Hardcoded 2 workers** — yes, wire it to the existing-but-unused `SstConfig.background_thread_count` at the spawn site (engine.rs:212). Default it to `min(2, cores/4)` so a small anchor node doesn't over-commit, mirroring the encode-pool headroom rule (TD-FLUSH-5).

### Co-design note
The 7× I/O claim is sound and the dominant-cost-term framing (GET round-trips) is correct. One addition: D2's stop-watermark backpressure must carry `TenantContext` so a stalled write is billed/attributed correctly (fail-closed) — the admission boundary is a metering boundary too (mandate #5). Worth a line in the boundary contract.

PR #1253, opened 2026-07-27
feat(sst): TD-COMPACT-6/7 — async compaction + L0 admission watermarks (ADR-076 D1+D2)

vjsingh1984:
Closes ADR-076 D1+D2 — the coupled producer/consumer rate-control loop (D0
unify-instances and D3 training-debounce landed in #1242 / #1241).

## Problem

The SST flush path ran compaction **inline-synchronously** (`run_due_compaction`,
awaited at `flush/mod.rs:747-781`), blocking every flush for ~35s of PCA training
CPU during ingest. The TD-COMPACT-5 training arm re-fired on every flush (no
debounce on the async boundary), so up to N−1 redundant training cycles per
ingest batch — each costing N object-store GETs + ~35s CPU + 1 PUT + N DELETEs.
ProximaDB was the outlier vs RocksDB/HBase/Cassandra (all async + watermarked).

## D1 — async compaction (flush → `schedule_compaction`)

- `Compaction::enqueue_due_compaction` mirrors `run_due_compaction` but enqueues
  to the background worker pool instead of performing inline. Flush returns ~1s
  after the L0 write (not ~35s). `run_due_compaction` retained as the
  inline/test primitive.
- **Race fix (cross-model review):** the `active_compactions` insert moved from
  worker-time to enqueue-time, under one write lock with the dedup check —
  closes the TOCTOU where two rapid schedules for the same output both passed
  the read check before either worker inserted.
- **Atomic-coordinator preservation:** the `SstEngine`'s `Compaction` is now
  built via `with_atomic_coordinator(Some(coordinator))` so the background
  workers carry the same ADR-046 atomic-swap coordinator the inline path had
  (`Compaction::new` passed `None` → non-atomic publish on the async path).
- **Worker state release:** `worker_loop` runs `release_task_state` on every
  exit path (Ok / Err / manager-construction failure), clearing the per-output
  active marker and the per-collection `training_in_flight` guard (collection
  dir derived from `task.output_file.parent()` — no new task field needed).
- `background_thread_count` wired to the worker spawn at both `Compaction`
  instances (was hardcoded `2`).

## D2 — L0 admission watermarks (TD-COMPACT-7)

- `SstConfig.l0_slowdown_trigger` (default **6**) + `l0_stop_trigger` (default
  **10** = `max_files_per_level`). Cross-model review: `stop ≤
  max_files_per_level` so the stop, not the level ceiling, is the active
  boundary. Both env-overridable (`PROXIMADB_L0_SLOWDOWN/STOP_TRIGGER`),
  TOML-controllable; `0` disables.
- Flush-path check before the L0 write: ≥ slowdown → yield ~1ms; ≥ stop →
  return `Err` so `materialize_collection`'s `?` returns before its `free_wal`
  block (data stays durable in the WAL; the auto-flush driver retries next
  tick). Armed only when a `compaction_manager` exists.

## Co-design note

The dominant cloud cost term is I/O round-trips + egress, not CPU. D1 removes
the inline ~35s CPU block (flush throughput ~30×); D2 bounds the L0 backlog so
async compaction cannot be starved into an unbounded GET storm. The
`training_in_flight` guard (D3, lifted to the worker here) eliminates the
redundant re-training loop — the single biggest COGS waste during ingest.

## Verification

- `cargo check --all-targets` ✓; `cargo clippy --lib --bins -- -D warnings` ✓;
  `make work-commit-check` ✓ (fmt, panic-policy, layering, tenant-path,
  capability-matrix, env-gates).
- New tests: `td_compact6_schedule_dedups_same_output_at_enqueue` (dedup at
  enqueue, no workers — deterministic) +
  `td_compact6_worker_clears_training_in_flight_after_completion` (async guard
  release via `await_compaction_quiescence`). Both pass.
- 71 compaction + 110 flush lib tests pass (no regressions).
- End-to-end ingest throughput + L0-bounded measurement on Azurite is the
  follow-up (this PR ships the mechanism + unit proof).

## Docs

TD-COMPACT-6/7/8 → Implemented; ENV_GATE_REGISTRY updated for the two new gates.


vjsingh1984:
## Blocker found (end-to-end verification) → fixed → verified ✅

The plan's "measure, don't assert" step caught a defect CI + unit tests both missed.

**The hang:** the async compaction worker built a *fresh* `Compaction` per task (`with_atomic_coordinator` → a cold, separately-built reader) and called `perform_compaction` on it. The inline path that worked used the persistent `self` (warmed at boot). The cold instance's read path **hung** — `perform_compaction` entered, never returned (0 completions over 12+ min; one thread at 100% CPU). With D2 on, L0 never drained → the stop watermark fired 530× → **ingest deadlock**. Unit tests missed it (nonexistent input → fast error, no real compaction).

**The fix (commit 2c47492):** `start_workers`/`stop` take `self: &Arc<Self>`; each worker holds an `Arc<Compaction>` clone of the persistent instance and runs `perform_compaction` on it directly. No per-task `temp_manager`.

**Consolidation (per the cleanup directive — net −44 LOC):**
- Removed the per-task `temp_manager` (the obsolete construction that hung).
- Removed dead `run_due_compaction` (zero callers).
- `enqueue_due_compaction` / `perform_compaction` dropped their unused config/coordinator params.
- `perform_unified_vectorrecord_compaction` reuses `self.filesystem_factory` (was one-per-call).
- StorageEngine's redundant `Compaction` no longer starts 4 idle never-fed workers.

**Verified (release binary, 150k + 300k SIFT, flush_interval=12s):**
- D2 off: compaction **completed in ~2 s**, 5 L0 → 1 L1 (drained). Was 0 completions.
- D2 on (defaults 6/10), 300k: **4 compactions completed, 0 D2-STOP (was 530)**, 0 slowdown, 3/3 flushes ok, 300k in 8.2 s.
- td_compact6 + 71 compaction + 110 flush lib tests pass; clippy `--lib --bins -D warnings` clean; work-commit-check clean.

**Known follow-up (not blocking):** the StorageEngine's `Compaction` instance is still created for `set_precision_resolver` (database.rs), but that resolver targets the idle instance, not the SstEngine's one that actually compacts — so fp16 precision-preservation through compaction is likely ineffective today. Separate PR.

PR #1325, opened 2026-07-30
feat(storage): reduce GETs with persistent cache and admitted flushes

vjsingh1984:
## Summary

- add persistent local-disk L2 tiers for PAX invariant and survivor caches, including spill, read-through, write-time population, metrics, configuration, and lifecycle invalidation
- admit flush work before sort, training, and encoding; preserve independent per-collection progress and typed WAL backpressure
- consolidate higher compaction levels, coalesce PAX probe ranges, and remove redundant PAX header/magic reads
- move internal collection identity toward stable numeric IDs and keep string/base62 encoding at external/path boundaries
- harden the SIFT benchmark harness with exact ingest accounting, provenance, settled-state, layout, recall, GET, and phase-scoped latency gates

## Checkpoint evidence

- release-server binary built from 25955abd83a61da9c9930fedb322d280c1b9c9bd
- focused typed-backpressure test passed
- production clippy passed with warnings denied
- rustfmt and repository env-gate checks passed
- SIFT1M file-backed cache-cold: recall@10 0.9812, 16.59 counted reads/query, 48.53 MB/query, p50 38.53 ms
- SIFT1M local-disk warm: recall@10 0.9804, 0.01 counted object-store reads/query

## Remaining before merge

- rebase the three current develop commits
- run the cold path through Azurite HTTP with persistent cache on actual local disk; file:// validates the read seam and counts but not Azure HTTP latency
- resolve and document the bounded single-level L0 tail versus trained two-level PAX layout acceptance rule
- update benchmark evidence and TD/ADR records from the final controlled run
- drive all required CI checks green


vjsingh1984:
Focused design-invariant review from the author of the handoff (#1302/#1304), ADR-080, and TD-COMPACT-10 — checking the implementation against the design intent + the mandate-#12 failure modes, not a line-by-line. Overall this lands the full D1–D6 stack and the measured evidence (cold 16.59 reads/q, warm 0.01, recall 0.98) is strong. A few things worth addressing before merge:

**1. GET-metering semantics changed in this PR — the before/after comparison is confounded.** `counting.rs` now counts each constituent range of a `read_ranges` batch as a separate GET (`physical_gets = ranges.len()`) instead of 1-per-batch. The comment *asserts* "`read_ranges` delegates to one `read_range` per constituent range," but doesn't prove the underlying object-store backend doesn't coalesce adjacent ranges into fewer wire GETs. Two asks:
  - Confirm `FileSystem::read_ranges` (and the object_store Azure/S3 backend beneath) actually issues N physical GETs — if it coalesces, the new count over-counts and 16.59 is inflated.
  - The 16.59 headline uses the new metering; prior baselines (e.g. the 48.5 in TD-COMPACT-10) used the old (1-per-batch). Provide a same-metering before/after (or re-measure develop under the new metering) so the reduction magnitude is verifiable. (Bench-only under `PROXIMADB_COUNT_FS_IO=1`, so this is measurement-accuracy, not production behavior — but it's the PR's headline number.)

**2. `nprobe_multiplier` default raised 1.0→2.0** (`src/core/config.rs:1457`, `CoarseProbeConfig`) — correct recall fix (1.0→nprobe 6→recall 0.9786 missed the ratchet; 2.0→nprobe 11→cleared it), but it raises per-query cell probes (~6→11 per segment) and therefore GET pressure that the L2+coalescing now has to offset. Worth a one-line note in the PR/ADR that the cache is partly compensating for a recall-driven nprobe bump, so future nprobe tuning and cache-hit-rate are understood as coupled.

**3. WAL-durability invariant — SOUND.** `persistent_l2.rs` is explicitly "deliberately a cache, never a correctness authority" (line 6), stores immutable cache bytes (not WAL), carries per-64 KiB chunk CRCs, and invalidates on compaction (`remove_where`, :266). Respects ADR-0050; no mandate-#4 panics. (Segments are uniquely-named, so even an orphaned entry is a disk-hygiene matter, not correctness — but worth confirming `remove_where` is actually wired at every segment-lifecycle event so the L2 doesn't grow unbounded.)

**4. Verify the 0%-hit fix mechanism.** The read-path `cache_changed` gating (`segment_format.rs:2391`) is structurally unchanged from the version that exhibited the 0% invariants-cache hit (the D2 prerequisite in #1304 Acceptance Criterion 0). The new `put_with_l2` is wired at write-time (D6, `:2717`) and read-through — so the cold-first-query hit may now come from write-time population rather than the read path. If so, confirm the invariants cache still populates on a cold first query when `cache_on_write = none` (i.e. the read path fills it independently), or document that the warm invariants hit depends on D6 being enabled. The warm=0.01 number is encouraging but doesn't isolate the invariants tier from the survivor tier.

**5. `level_multiplier` stays 1.0 + bounded [1.0, 1.5]** (`config.rs:1659`) — consistent with ADR-080's amended recommendation. Good.

Net: the architecture is sound and the measured GET reduction is real under the new metering; #1 (metering comparability) and #4 (0%-hit mechanism) are the two I'd want resolved before this becomes the GET-reduction baseline.

vjsingh1984:
Follow-up on my Finding #1 above — I verified the metering question and need to correct the record (in this PR's favor).

**The new per-range metering is CORRECT, not an over-count.** Verified two layers:
- **Cloud billing**: Azure Blob, S3, and GCS all bill 1 GET per ranged request, and none support multiple disjoint ranges in a single GET — each range is its own HTTP request. So N non-contiguous ranges = N billed GETs on every cloud.
- **Impl**: the `read_ranges` default (`crates/storage/proximadb-storage-filesystem-types/src/lib.rs:89-90`) fans out to exactly 1 `read_range` per range, and **no cloud backend overrides `read_ranges`** to coalesce — `azure_blob.rs` / `aws_s3.rs` / `gcs_store.rs` implement only `read_range`. Range coalescing happens higher (PAX scanner `plan_probe_cell_ranges`), *before* `read_ranges`, so the ranges reaching the metered layer are already final. `ranges.len()` = real billed GETs.

So my earlier "backend may coalesce → 16.59 could be inflated" worry was wrong. The metering is exact and **cloud-agnostic** — the only per-cloud difference is the $/op rate applied after counting (ADR-0043), not the count itself.

**The comparison-confound concern flips direction:** the *old* 1-per-batch metering **under-counted** (a 5-range batch was counted as 1 GET when the real bill was 5). So the prior 48.5 baseline was too *low*; the new 16.59 is accurate; the real GET reduction is **larger than 48.5→16.59 implies** (this PR understates its own win). The one ask that remains: re-measure develop under the new per-range metering so the headline is an apples-to-apples same-metering delta.

(Finding #4 — verify the 0%-hit fix mechanism on the cold-first-query read path — still stands.)

vjsingh1984:
For the metering-correctness thread above: I grounded the underlying premise (all three object stores bill 1 read op per ranged GET; none support multi-range GET) in primary sources and filed it as a References section on anvaiops ADR-0043 (the per-cloud read-cost ADR) — anvaiops/anvaiops#427. Each cloud is cited from its primary API doc (single-range-per-GET) + pricing page (per-operation billing):

- **S3** GetObject: *"Amazon S3 doesn't support retrieving multiple ranges of data per GET request."*
- **Azure** Get Blob: single `Range` header + Billing section → Read operations per transaction.
- **GCS**: Class B per-operation billing (reads); single Range per GET.

Net: counting `ranges.len()` is exact and cloud-agnostic — only the $/op multiplier is per-cloud (ADR-0043's table). So the new per-range metering here is correct; the before/after comparability note from my earlier comment still stands (re-measure develop under the new metering for an apples-to-apples delta).

vjsingh1984:
📨 **Handoff for Codex** (taking over resumed after credit exhaustion). Full diagnosis + partial fixes in `docs/12-design/HANDOFF_PR1325_CODEX_HANDOFF.md` (just pushed, commit c952ae36a).

## What I fixed (committed; build on or rework)
1. **Storage CI failure** — bare-`SstEngine` tests used string ids vs the PR's fail-closed decimal `CollectionObjectId` requirement. Catalog-style numeric-id mock. **Storage tests now green.** (Rework to composite-keyed per ADR-0083 rev2 when that migration lands.)
2. **Flush-plan regression** — `CollectionFlushPlan.collection_object_id = coll.id.parse::<u64>()` errored on the UUID `coll.id`. Fixed per ADR-0083 rev2 **D2**: `get_collection_object_id` is now **parse-or-deterministic-hash** (a *derived, non-stored handle* — `stable_collection_handle`), same in the embedded flush-plan construction. Unblocks the flush; compiles clean.

## What remains — the graph/fusion CI failure (diagnosed)
`unified_search_native_with_tenant_context` returns **`Ok([])`** for the SST engine, while `VectorSearchRequest` (what `db.search`/test_large_k use) **works**. **`unified_search_native` is unchanged by this PR** — so the break is **downstream** in the SST engine: the PR's SST changes (persistent L2 / coalescing / compaction) broke the **native-search read path**, while the request-search path still works.

**Most likely culprit:** the new persistent L2 cache serves empty/stale to the native-read path (VectorSearchRequest likely doesn't route through it).

**Ruled out:** engine mismatch (both sst), search-path identity fail-close (none), `create_graph` flushing `code_vecs` (it doesn't), error swallowing (`retry_vector_search` bails, doesn't swallow).

**Where to look:** trace `unified_search_native` → SST native-search; **diff against the `VectorSearchRequest` SST dispatch** — the divergence (likely the L2 on the native path) is the bug. It's your SST/L2 code, so you'll recognize it fastest.

## Also (not CI-gating, but real): restart-recovery
`embedded_flush_recovery` loses everything after reopen — segments orphaned because the segment path is keyed by `coll.id` (UUID, re-minted each create). Fix per ADR-0083 rev2: key the path by the stable composite. (Confirm whether this suite is CI-gated — it wasn't in the original failure set.)

## Context
- **ADR-0083 rev2** (PR #1329 → develop): composite `CollectionIdentity` is the sole identity; global u64 retired; supersedes ADR-031. The unstable-identity (UUID `coll.id`) is the common thread.
- My commits: `8fcc57276` (partial fixes) + `c952ae36a` (this handoff doc).

The valuable part is the **localization** (native-search SST path empty; request path works; suspect persistent L2) — should save you the debugging.

vjsingh1984:
Resolved the remaining graph/fusion blocker in d3cafa02f.

Root cause: unified_search_native_with_tenant_context converted the catalog CollectionObjectId handle to a string and passed it into the WAL/memtable/SST search path as though it were the collection storage key. Fusion vectors were still memtable-resident, so native search selected an empty namespace and returned Ok([]); persistent L2 was not involved.

Fix: retain catalog-handle use for validation/indexing, but delegate native search with the tenant-resolved collection id or the original single-tenant collection key.

Local evidence:
- embedded_code_graph_parity_e2e: before 2 passed / 2 failed; after 4 passed / 0 failed
- sst_compaction_execution_test: 2 passed
- sst_ivf2_compaction_route_proof_test: 2 passed
- cargo clippy -p proximadb --lib -- -D warnings: passed
- cargo fmt --all -- --check: passed

PR #1345, opened 2026-07-31
bench: make PAX scale geometry evidence auditable

vjsingh1984:
## Summary

- add BIGANN u8bin and exact prefix-ground-truth support for 100K through 10M scale evidence
- stream controlled bulk ingest through Arrow Flight UPSERT while retaining canonical REST v2 search semantics
- make explicit per-collection flush synchronously materialize and retire the admitted WAL epoch
- make paginated v2 scans storage-inclusive after flush and pin real SST/PAX post-flush behavior
- isolate explicit flush epochs from timer races and validate Flight acknowledgements

## First-principles measurement contract

The query matrix reuses one immutable settled PAX bed per corpus size. Every nprobe/top-k point starts a fresh release server, uses exact corpus-scoped ground truth, and records persisted cell geometry, physical GETs/bytes, cache tiers, probed rows, latency, and recall. Azure/Azurite is identified separately from local-file evidence.

## Validation completed

- `cargo test --release --test v2_post_flush_search_e2e -- --nocapture --test-threads=1`
- `cargo clippy --release -p proximadb --lib -- -D warnings`
- focused Python harness tests: 18 passed
- Ruff check/format: clean

The five-point 100K/330K/1M/3.3M/10M matrix, fitted knee analysis, and SVG evidence will be added on this PR after the release-server build and controlled runs complete.


vjsingh1984:
Rebased head 04f691e7e onto develop and revalidated locally: benchmark/analyzer Python tests 26/26; admitted flush materializer 14/14; selective WAL claims 3/3; vector_ann_recall_under_deletes_and_compaction recall=1.000 with zero leaks pre-flush, post-flush, and delete-after-flush/reflush; cargo fmt clean; exact CI clippy command clean. The prior SIFT CI red was a Hugging Face CDN timeout before execution. The query red exposed batch-granular retirement of tombstone-only sources; the fix now publishes only vector-only batches, retains deferred deletion intents, and fails closed on mixed batches. cold-deletion-vectors remains default-off and is not yet consumed by PAX search, so it cannot safely replace WAL deletion intent yet.

PR #1414, opened 2026-08-03
fix(abac): carry the policy key to the relational enforcement point; SQL ports take PortIdentity (TD-ABAC-10b)

vjsingh1984:
## The bug

`DmlService::abac_row_filter` reads `tenant_stable_id` off the storage `TenantContext` — but the relational pipeline built that context with `TenantContext::for_tenant_id()`, which **always leaves it `None`**, and the filter treats an absent policy key as `Denied`.

With `abac-policy` enabled and ABAC stores provisioned (which *does* arm the enforcer in `SharedServices`), **every governed relational SELECT returned zero rows.** The relational boundary could not admit anyone — TD-ABAC-3's subject plumbing was effectively inert in production.

**The symmetry is the interesting part:** one root cause — the policy key never reaches the enforcement point — produced *opposite* failure modes on the two boundaries: **silent-open** on the records/vector seam (ADR-087's P1), **deny-everything** here. That divergence is precisely what ADR-087's "resolve once, carry unchanged" decision exists to remove, and it's why this is a functional fix, not just preparation for TD-ABAC-11's flip.

## The fix

**Carry the key end-to-end**
- **pgwire**: `session_tenant_stable_id()` → `try_run_select` → the snapshot's `TenantContext` → `scan_table_relational`. The EXPLAIN chain (`explain_*_with_catalog` → `route_and_plan_select` → `prepare_select_plan` → `build_snapshot`) carries it too, so `EXPLAIN ANALYZE` *executes* under the same decision it displays.
- **SQL ports**: `ApiHandlersPort::execute_sql_v1` and `QueryAdapterPort::execute_sql` widened from the flat `tenant_id`/`subject` pair to **`PortIdentity`** — the same consolidation TD-ABAC-7 applied to the vector ports, rather than bolting on a third loose parameter. gRPC SQL now supplies the key from `grpc_auth::tenant_stable_id` (its *records* paths already carried it); the api-crate hybrid SQL route projects the foundation identity.
- **Honest anonymity**: embedded (single-process, genuinely no per-request identity) and the unrouted root REST `execute_sql` handler (dead code) pass `PortIdentity::anonymous()` rather than being given an invented identity source.

**Test**: `RecordingApiPort` now records and asserts `tenant_stable_id`, so the port test fails if the key stops arriving — the exact regression that caused this bug.

## Verification

- 16/16 ABAC integration tests ✅ · 205/205 runtime+api lib tests ✅
- `cargo check --workspace --lib` (default features) ✅ — the embedded binding was the last caller
- clippy: root lib `--features abac-policy` ✅, leaf crates `--all-targets` strict ✅
- `cargo fmt --check` ✅ · `make work-commit-check` ✅ · `td-adr-unique` (87 ADRs / 331 TDs) ✅

TD-ABAC-10 is now **Resolved** (10a constructor + 10b carrier/bug + 10c the three bypasses).

## Next

TD-ABAC-11 only: default-tenant mint → flip the composition rule to fail-closed on an absent key (now that the key actually arrives, the flip is safe *and* meaningful) → cross-surface e2e ratchet. Note the flip must cover **both** composition points — `records_read_context` and `scan_table_relational` — as recorded in the TD-ABAC-11 addendum.

### DIFF
--- a/crates/binding/proximadb-embedded/src/lib.rs
+++ b/crates/binding/proximadb-embedded/src/lib.rs
@@ -4394,8 +4394,9 @@ impl EmbeddedProximaDB {
                             .collect()
                     }),
                     collection.map(str::to_string),
-                    None, // embedded single-process: no per-request tenant
-                    None, // TD-ABAC-5: embedded single-process: no per-request subject
+                    // Embedded is single-process with no per-request tenant or
+                    // subject, so there is no identity to carry (TD-ABAC-10b).
+                    proximadb_runtime::PortIdentity::anonymous(),
                 )
                 .await
                 .map_err(|e| -> Box<dyn std::error::Error + Send + Sync> {
--- a/crates/platform/proximadb-api/src/lib.rs
+++ b/crates/platform/proximadb-api/src/lib.rs
@@ -120,6 +120,7 @@ pub(crate) mod test_support {
             collection: Option<String>,
             tenant_id: Option<String>,
             subject: Option<String>,
+            tenant_stable_id: Option<u64>,
         },
         Hybrid,
     }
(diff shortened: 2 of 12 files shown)


vjsingh1984:
Closing as **superseded by concurrent work** — not merging a duplicate.

While this branch was in CI, #1405 (`carry canonical identity through relational reads`) and #1403 (`make tenant stable IDs total and fail closed`) landed on develop and implement the same slice. I verified develop's version before touching my conflicts, and it is **strictly better** than what this PR proposed:

- The composition rule now lives in the **foundation crate** as `read_enforcement_composition` with a full 24-case truth table — one definition shared by the vector seam *and* the relational boundary, rather than my per-boundary threading.
- It adds `OwnedPortIdentity` (`into_owned`/`as_borrowed`) so planners can retain the identity past the protocol call frame — a carrier form this PR lacked, and the cleaner answer for `build_snapshot`/`abac_row_filter`, which now take `PortIdentity` directly instead of parallel params.
- Everything this PR added is covered: the relational + EXPLAIN threading, both SQL ports on `PortIdentity`, gRPC SQL's policy key, and the `RecordingApiPort` `tenant_stable_id` assertion.

The underlying deny-everything bug this PR was written to fix (absent policy key ⇒ `abac_row_filter` denies ⇒ every governed relational SELECT returned zero rows) is fixed on develop by the same mechanism.

Worth noting: the known-open-cell test I pinned in #1400 did its job — landing the fail-closed flip required consciously renaming `composition_rule_still_passes_through_a_subject_without_a_stable_id` → `composition_rule_denies_a_subject_without_a_stable_id` and inverting its assertion. The behavior could not change silently.

PR #1422, opened 2026-08-04
fix(storage): publish spill outputs once across backends

vjsingh1984:
## Summary

- preserve the catalog-owned numeric collection identity while keeping physical base62 paths behind the default-OFF typed-path gate
- publish disk-backed spill compaction output once to a unique final object key through bounded Azure/S3 multipart or GCS resumable upload
- retain same-directory hidden staging plus atomic rename for file and bare local paths, including embedded mode without a transaction coordinator
- reject unsupported remote publication backends before sorting, training, or PAX construction
- document the 10M Azurite publication failure as diagnostic evidence and require the release query matrix to be rerun on the corrected path

## Verification

- storage-ports unit tests: 16 passed
- GCS resumable planning/range tests: 2 passed
- SST staged-write atomic visibility/cleanup tests: 4 passed
- forced local-spill PAX/MVCC integration: passed
- cloud-full object-store contract target: builds cleanly
- Azure/Azurite 17 MiB+19 multipart round trip: passed
- S3/MinIO 17 MiB+19 multipart round trip: passed
- GCS/fake-GCS 17 MiB+19 resumable round trip: passed (308, 308, 200)
- formatting, env-gate registry, diff whitespace, and TD/ADR uniqueness checks: passed

## Promotion notes

- zero behavior change when PROXIMADB_TYPED_PATHS is unset
- source segments and deletion vectors remain authoritative until final-object publication succeeds
- the 10M recall/GET/latency matrix remains an evidence gate and is not inferred from the earlier failed publication run

### DIFF
--- a/apps/proximadb-server/tests/object_store_restart_recovery.rs
+++ b/apps/proximadb-server/tests/object_store_restart_recovery.rs
@@ -11,7 +11,6 @@
 //! by default because it requires cloud credentials and
 //! `PROXIMADB_OBJECT_STORE_URL=s3://...` or `adls://...`.
 
-use std::net::TcpListener;
 use std::path::Path;
 use std::process::{Child, Command, Stdio};
 use std::time::{Duration, Instant};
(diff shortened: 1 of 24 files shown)


vjsingh1984:
Follow-up CI correction pushed in 955990254: the flush boundary now accepts only decimal catalog object IDs and no longer hashes aliases/UUIDs into a second identity domain. Verified cargo test -p proximadb-storage-traits --lib (23 passed) and embedded_flush_recovery embedded_flush_persists (2 passed). The exact release-server Azure build is now running from this clean commit.

PR #1514, opened 2026-08-06
feat(authz): L0.1 — catalog-resident principal + API-key registry (ADR-090)

vjsingh1984:
First implementation slice of **ADR-090** (#1513), tracked by **TD-AUTHZ-1 L0.1**. TDD: the seven tests in the module *are* the specification — written to pin the contract, each named for the clause it pins.

## What

`proximadb-catalog::principal_registry` (behind the existing `fc-metamodel` feature — it shares `SubjectId` with the ABAC metamodel):

- **`PrincipalRecord`** — a user; `tenant_id` **required** (ADR-090 L0: tenant derives from the credential, never a client-asserted header). `(tenant, subject)` is the key — "alice" in two tenants is two principals.
- **Mint / revoke / resolve / list** over `pxk_<key_id>.<secret>` (8-byte public id, 32-byte OS-random secret). **Only the SHA-256 of the secret is stored**; the mint return is the only copy that ever exists — and a test reads the persisted snapshot back to assert the secret never touched disk. SHA-256 rather than a password KDF is deliberate and documented: these are 256-bit machine secrets, not passwords.
- **`resolve_key` is fail-closed and uniform**: malformed / unknown id / wrong secret / revoked / expired / principal disabled or missing all return `None` with no distinguishing detail; the secret compare is constant-time.
- **Disable-principal kills all their keys at resolve time**; rotation is mint+revoke with a graceful window (tested).
- Persistence copies the ABAC stores' OSS mechanism exactly: JSON snapshot, atomic tmp+rename with fsync, load-on-open. Reopen round-trip tested, including revocation durability.

## Gaps closed (ADR-090 appendix A)

| gap | status |
|---|---|
| 1 — no key store (config.toml, restart to add) | ✅ durable registry, runtime mint/revoke |
| 2 — plaintext keys, no constant-time compare | ✅ hash-only at rest, ct compare |
| 4 — `tenant_id` optional on keys | ✅ required, enforced with a test |
| 6 — no user registry / tenant→users enumeration | ✅ `list_principals` |
| 3 — three competing key tables | follow-on: `SecurityCoordinator` adopts this registry; config table becomes a bootstrap import |

## Design notes

- **Modality-agnostic by construction**: principals/keys are catalog objects that refuse to know what a collection is. Multi-modal authorization comes from L1 grants scoping over the collection abstraction — which is what keeps identity uniformly true for vector, graph, document, timeseries, and relational.
- **Mechanism-only per the OSS boundary**: who *may* mint, tier key-count limits, seat billing = commercial policy, not this crate.
- Panic-policy: no unwrap/expect on production paths (poisoned-lock recovery via `into_inner`).

## Verification

7/7 spec tests green (0.20 s) · `cargo clippy -p proximadb-catalog --features fc-metamodel --all-targets -- -D warnings` clean · dependent `proximadb-abac` compiles · `make work-commit-check` green.

Ref ADR-090 L0.1, TD-AUTHZ-1, #1513.


vjsingh1984:
Pushed the second half of the slice: **L0.1 wiring** (`a06e750c2`) — `UnifiedAuthService` adopts the registry.

**The rule:** the `pxk_` namespace is **registry-authoritative**. A credential that looks like a registry key never falls through to the config table — no registry attached, wrong secret, revoked, expired, disabled principal all fail closed with one uniform error. A config entry mimicking the registry format is a spoof attempt, not an alternate authority (pinned by test in both the empty-registry and no-registry cases).

**Registry identities carry empty permissions, deliberately** — the credential proves *who you are*; *what you may do* is grants/ABAC (L1/L2), not strings beside a key.

The seam is default-`None`: nothing changes for any deployment until the boot wiring lands (next slice: open at `<data_dir>/identity` in `SecurityCoordinator`, config table becomes a bootstrap import).

Verified: `cargo test --lib security::auth_service` **16/16** (2 new + 14 pre-existing, no regressions), `cargo check --lib` clean, `make work-commit-check` green.

vjsingh1984:
Third commit: **L0.1 boot wiring** (`06a7440f9`) — `initialize_security_with_identity` opens the registry at `<data_dir>/identity` (beside `<data_dir>/abac`) and attaches it before the Arc wrap. **Fail-closed boot**: a security-enabled deployment that can't open its identity store aborts. Old `initialize_security` stays as a back-compat delegate.

The L0.1 arc is complete: registry → service seam → live boot path. A `pxk_` key now authenticates end-to-end.

Gate note: `cargo check --lib` would have missed the two changed `#[cfg(test)]` call sites — verified on the lib **test** profile instead (`cargo test --lib database::` exit 0).

Issue #1527, opened 2026-08-07
record_count reports 0 for a populated collection (scan and point lookup both return its contents)


vjsingh1984:
## Summary

A collection holding 1,000 flushed records reports `record_count: 0` and `storage_size_bytes: 0`. Both retrieval paths disagree with it — `scan` returns all 1,000 and point lookup returns individual records with vectors and props intact.

Verified on `develop` @ `df8ece037`.

## Reproduction

Create an SST collection, insert 1,000 records, restart the server (so the data is flushed to an L0 `.pax` segment rather than sitting in the memtable), then:

```
POST /api/v2/collections/t/records/scan  {"limit":5000}
  → 1000 records

GET  /api/v2/collections/t/records/r5
  → {"id":"r5","vector":[0.0353,0.9608,...],"props":{"f":{"type":"string","value":"a.py"}}}

GET  /api/v2/collections/t
  → "stats":{"record_count":0,"storage_size_bytes":0,"indexed_fields":0,"text_field_count":0}

GET  /api/v2/collections
  → "record_count": null
```

On disk at that moment: one `.pax` segment holding the 1,000 records.

## Why this is worth fixing rather than documenting

`null` in the list response is defensible — "not computed" is an honest answer. `record_count: 0` next to `storage_size_bytes: 0` is not: it is a **confident wrong answer**, and it is indistinguishable from a genuinely empty collection.

That distinction has a real cost. This session I used exactly these surfaces to judge whether data had survived a restart, concluded a healthy database had lost everything, and filed a durability bug that turned out to be false. A store that answered "unknown" would have prompted me to check; a store that answered "0" convinced me. It also silently invalidates capacity/footprint tooling built on collection stats.

## Expected

Either report the real count, or report it as unknown (`null` / omitted) consistently across `GET /collections/{id}` and `GET /collections`. Anything that cannot be computed cheaply should not be defaulted to `0`.

If a cheap exact count is not available for flushed segments, a segment-metadata-derived estimate with an explicit `approximate: true` flag would be more useful than either current answer.

## Related

- #1524 — the `scan` half of this observability gap; already fixed on `develop`, which is how the count divergence became visible in isolation.

vjsingh1984:
## Root cause: the catalog counter exists and is never incremented

The machinery is all present. `MetadataStore::update_stats` maintains the counter by delta (`storage/metadata/write_ahead_log.rs:596`):

```rust
if vector_delta > 0 {
[...]

Issue #1524, opened 2026-08-07
bug(embedded): ORION edges and ProximaRecords do not survive a restart


vjsingh1984:
Two durability defects in embedded mode. Both write paths report success and read back correctly **in the same session**; the data is gone after the process restarts. Filing above #1479 (write amplification) because that one is an optimisation and this is correctness.

## 1. ProximaRecords are not persisted

```
POST /api/v2/collections/t/records/batch
  -> {"inserted_count": 200, "failed_count": 0, "errors": []}

POST /api/v2/collections/t/records/scan   (same session)  -> 200 records
stop server; on-disk data dir                             ->  11 KB
POST /api/v2/collections/t/records/scan   (after restart) ->   0 records
```

Not a collection-config issue — identical behaviour with and without `enable_proxima_record: true` (tested both). `create_collection` in `embedded.py` never sends that flag, so collections default to `proxima_record_enabled: false`, but forcing it true changes nothing.

There is also no flush/checkpoint API for record collections, although `document.py:1520` and `timeseries.py:1432` both expose `flush()`.

## 2. ORION loses edges across a restart (but keeps nodes)

Same corpus, same process, via Victor's graph store:

```
before close:  7 nodes,  7 edges
after reopen:  7 nodes,  0 edges
```

Nodes survive; edges do not. **This appears to be flush-threshold dependent** — a larger `victor init`-scale write (1,328 nodes / 1,258 CONTAINS edges) *did* retain its edges across a restart, while small writes lose them consistently. Durability that varies with write size is harder to trust than a flat failure, since it looks fine at the scale people demo with.

## Why this matters downstream

Victor's design (TD-12) makes the ProximaRecord the authority for a code symbol — graph properties **and** its embedding in one record, with ORION as a rebuildable projection. With records non-durable, that authority does not exist on disk: a repo indexed in one process comes back with no vectors and a partial graph, silently. It also blocks moving statement-level CPG fragments into ProximaDB, since that would mean losing the bulk of the graph on every restart.

Two stubs make this easy to miss: `count()` and the collection's `record_count` both answer `0` regardless of contents, so an empty collection is indistinguishable from a full one unless you scan.

## Ask

Either flush-on-stop plus WAL replay on start, or an explicit record-collection `flush()` mirroring the document/timeseries ones — plus whatever makes ORION edge persistence independent of write size.

## Reproduction

[...]

PR #1528, opened 2026-08-07
fix(storage): surface segment-listing failures instead of reporting an empty collection

vjsingh1984:
## Problem

`read_all_records` discards the error from `fs.list` on both the SST and HELIX engines:

```rust
if let Ok(entries) = fs.list(storage_url).await { ... }
if files.is_empty() { return Ok(Vec::new()); }
```

A segment directory that cannot be listed therefore produces exactly the same observable result as a collection that genuinely has no data: an empty vec, no log line, no error, no distinction available to any caller.

Every consumer inherits the ambiguity. `scan` reports zero rows. The TD-112 AXIS-rebuild-from-SST path (the comment at `trait_impl.rs:417` calls this out as the path that must read `.pax` "or the index stays empty after a loss") would rebuild an empty index. Neither surfaces anything.

## Why it's worth closing

This is not hypothetical. While investigating #1524 I used `scan` as a durability oracle across a restart, got zero, and concluded a healthy database had lost its records — twice, in opposite directions, before a point-lookup probe showed the data was intact all along. A store that had said "I could not list the segment directory" would have ended that in one step.

The general form: **"empty" and "I could not look" must not be the same answer.** An empty result is a legitimate, common outcome, which is precisely why it makes such a poor carrier for a failure signal.

## Change

- SST and HELIX `read_all_records` now `match` on the listing result and `warn!` on `Err`, naming the URL, the underlying error, and the collection. The genuinely-empty case stays silent — it is normal and should not be noisy.
- `list_all_records_with_tenant_context` logs the WAL and storage row counts it merges, at debug. A zero from either half alone is unremarkable (an empty WAL is expected after a flush; an empty storage read is expected for a new collection), but together they are the "this collection looks empty" signal, and there was previously nothing to inspect when that was wrong.

No behavioural change to the success path, and no change to what callers receive — this only makes an existing silent failure observable.

## Scope

Deliberately narrow. It does not attempt to fix `record_count` (#1527), and the scan defect that prompted the investigation turned out to be **already fixed** on `develop` — my repro was against a two-week-old binary, and I've corrected #1524 accordingly. What survives that correction is this silent-failure path, which is independent of the bug that led me to it.

## Verification

`cargo check -p proximadb-server` clean. The warning path is by nature not reachable from a healthy fixture; the change is the error branch of an existing `if let Ok`, with the success branch preserved verbatim.

### DIFF
--- a/src/services/operations/vectors/legacy.rs
+++ b/src/services/operations/vectors/legacy.rs
@@ -1312,6 +1312,17 @@ impl VectorOperationsService {
         let storage_records = engine
             .read_all_records(&collection_id, Some(&storage_url))
             .await?;
+        // Both halves are logged because a zero from either is silent by nature: an
+        // empty WAL is normal after a flush, and an empty storage read is normal for a
+        // new collection — but together they mean "this collection looks empty", and
+        // when that is wrong there is otherwise nothing to inspect.
+        tracing::debug!(
+            "list_all_records: collection={} storage_url={} wal_records={} storage_records={}",
+            collection_id,
+            storage_url,
+            wal_records.len(),
+            storage_records.len()
+        );
 
         // Merge by oid: storage baseline, then WAL overrides on >= updated_at_ns
         // so the freshest version (WAL) wins ties.
--- a/src/storage/engines/helix/mod.rs
+++ b/src/storage/engines/helix/mod.rs
@@ -2217,16 +2217,30 @@ impl UnifiedStorageFormat for HelixEngine {
             return Ok(Vec::new());
         };
         let fs = self.filesystem_for_url(storage_url, collection_id)?;
+        // Mirrors the SST engine: a listing failure is reported, not folded into an
+        // empty result that is indistinguishable from an empty collection.
         let mut files = Vec::new();
-        if let Ok(entries) = fs.list(storage_url).await {
-            for entry in &entries {
-                if !entry.metadata.is_directory
-                    && entry
-                        .url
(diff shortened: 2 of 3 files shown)


vjsingh1984:
## CI note: the SIFT ratchet is timing out on PR branches, and it is not this change

This PR has had `SIFT PAX recall ratchet (N=10k)` fail twice. Both were **cancellations during "Build ratchet binary"**, not recall regressions — the ratchet step itself never ran:

```
conclusion: cancelled
  Build ratchet binary                                     cancelled
  Run SIFT PAX recall ratchet (N=10000, floor=0.85)        skipped
```

Both attempts died at ~30 minutes (30m18s, 30m16s) against the job's `timeout-minutes: 30`. The same job on a **develop push** completed in 24m56s (run `31158464352`).

The difference is cache warmth: develop pushes reuse a warm `Swatinem/rust-cache`, PR branches build colder. That puts the job's build phase at roughly 25–30+ minutes against a 30-minute cap — it passes or fails on which side of the line the runner lands.

This change adds two `tracing` calls and converts an `if let Ok` into a `match`. It cannot plausibly account for 30 minutes of build time.

Worth raising separately: a ratchet whose build consumes its entire budget before the assertion runs is not gating anything on PR branches — it is a coin flip that reports as a recall failure. Either raising `timeout-minutes`, or splitting the build from the assertion so the timeout applies to the measurement rather than to compilation, would make the signal mean what its name says. Happy to file that if it is not already tracked.

PR #1536, opened 2026-08-07
fix(security): arm the security controls that existed but never fired (TD-SEC-2 Slice A)

vjsingh1984:
A pressure-test of the authn/authz/audit surface against common multi-tenant SaaS practice found that the headline problem is **not missing controls — it is controls that exist, look implemented, and never fire.** That is worse than absence: it manufactures false assurance.

## Five inert controls, now armed

| # | control | what was actually wrong |
|---|---|---|
| 1 | **Authentication audit** | `set_audit_logger` had **zero callers** → `audit_logger` always `None` → the emit hook in `authenticate()` was dead. **No auth success or failure has ever been persisted.** |
| 2 | **Failed-auth attribution** | Failures were attributed to `anonymous()`, and the detector keys on `user_id` → **every failure in the system shared ONE bucket**. An account under attack was indistinguishable from noise. |
| 3 | **Brute-force counter** | `query_events` truncated to `limit` *before* sorting → a bounded query returned an **arbitrary** subset, not the newest — silently defeating a threshold that samples the newest 100. |
| 4 | **Cross-tenant alert** | Compared `event.tenant_id` **with itself** → unsatisfiable → could never fire. |
| 5 | **Failed-auth throttle** | `RateLimitingService` fully implemented, **zero call sites**. |

**I initially claimed wiring #1 alone would arm the brute-force detector. That was wrong** — #2 and #3 would have left it broken, and the PR includes both.

## Design points worth review

- **Attribution is bound by a secrecy rule.** A `pxk_` key contributes its *public* key id (before the `.`); every other credential shape contributes a truncated SHA-256 **fingerprint** — stable enough to count against, useless to replay. Secret material must never reach an audit record.
- **The throttle gets its own budget knob** (`failed_auth_per_minute_per_principal`, default 10), not the request-rate knob — different semantics; conflating them makes one wrong. It counts **only** the principal dimension: the client IP isn't available at this seam, and a placeholder would recreate defect #2 (one global bucket). Real IP threading is the follow-on.
- **`AuditResource.resource_tenant_id`** is `#[serde(default)]` so records written before today still parse (mixed-read-safe).

## Two defects fixed

- **Boot footgun**: `security.enabled=true` + the **shipped default** `enable_audit_logging=false` **refused to boot**. Now mode-gated: Development gets a documented no-op sink + warning; every other mode still fails fast. Written `!matches!(mode, Development)` rather than `matches!(mode, Production)` **on purpose** — Enterprise is production-grade too, and any mode added later must fail **closed**.
- **Audit-failure handling**: the silent `let _ =` becomes `error!` + an `AUDIT_WRITE_FAILURES` counter; **default fail-open**, opt-in `audit_fail_closed`. Rationale is in-code so it can be argued with: the credible failure is a full disk (retention has no production caller), and denying on it is a self-inflicted total-auth outage; an adversary who can break the sink can already rewrite the unchained history. What matters is that the loss of accountability is **visible**. Separately, on the authz-denial path an audit failure used to **replace** "Insufficient permissions" — the denial now always stands.

## Correction to the record
A prior review note said no rate limiting exists on the request path. There **is** one: an opt-in per-IP limiter on **pgwire** (`PROXIMADB_PGWIRE_RATE_LIMIT_RPM`, default off). Nothing throttles REST/gRPC auth, and nothing counted auth *failures* in any transport — this adds the latter.

## Verification
5/5 new spec tests green — attribution never leaks the secret · distinct credentials don't share a bucket · budget is per-principal and a bystander is unaffected · a bounded query returns the **newest** events. Default build clean · `make work-commit-check` green.

Deferred to TD-SEC-2 (design already worked out, incl. the tamper classes it could *never* detect): the tamper-evident hash-chained audit ledger.


vjsingh1984:
Pushed the fix for both CI failures — both were mine, and both had the same root cause: **verifying with a gate that cannot see the code the change affects.**

1. **Rust Compile (test)** — adding `audit_fail_closed` to `AuthenticationConfig` is a signature change, and two integration tests construct it. `cargo check --lib` / `cargo test --lib` don't compile `tests/*.rs` (separate crates), so my local run passed while all-targets failed. Correct gate: `cargo check --all-targets`.
2. **Rust Clippy** — `AuditResource` imported at module scope but used only in `#[cfg(test)]` ⇒ unused in the non-test build ⇒ fatal under `-D warnings`. Moved into the test module.

Re-verified with CI's **exact** commands (`clippy --lib --bins -- -D warnings`, `check --all-targets`) rather than approximations — an earlier `--all-features` guess produced a misleading failure in an unrelated crate.

PR #1558, opened 2026-08-09
bench+td: coarse-PCA projection width is mis-calibrated (TD-IVF-3, TD-COMPACT-12) + harness fixes

vjsingh1984:
## Purpose

Files two TDs from a 16-bed measurement study of the IVF coarse-PCA projection width, plus the four harness fixes that study needed. Docs and bench-harness only — **no engine or product code changes**.

## TD-IVF-3 — projection width is mis-calibrated

`ivf_projection_dims()` gives **10 components for 128-d** and **14 for 768-d** (7.8% and 1.8% of the space). Coarse cells are *ranked* in that projection, so a cell that fails to rank inside `nprobe` is never read regardless of centroid precision.

| corpus | default | optimum | floor nprobe | GET/query | write |
|---|---|---|---|---|---|
| 128-d BIGANN 3.3M (k_c=100) | 10 | **32** | 16 → 12 | 33.4 → **26.3** (−21%) | +15% |
| 768-d BGE 491k (k_c=90) | 14 | **64** | 30 → 25 | 48.6 → **38.7** (−20%) | +8.9% |

Both optima beat every other width measured at their scale, on every shared point.

**The TD proposes a calibrated table, not a formula** — deliberately. Three a-priori framings were each fitted at 128-d and each failed out-of-sample at 768-d:

| framing | predicted 768-d optimum | actual |
|---|---|---|
| samples/param ≥ ~15 | 32 | wrong — 64 beat 32 (11/11 points) |
| fraction of dims ≈ 25% | ~190 | wrong — flat from 64 to 128 |
| variance target 79% | 163 | wrong — flat well below it |

What survives is two measurements and a sublinear interpolation (`n_comp ≈ 4.9·dim^0.387`), offered explicitly as an interpolation with its 384-d prediction of ~49 as the falsifiable test. Ships **default-OFF** behind the existing `PROXIMADB_IVF_NCOMP`.

## TD-COMPACT-12 — budget-refused compaction is silent

Not a duplicate of TD-COMPACT-11, which covers the stall itself and builds the spill executor. TD-COMPACT-11's exit criteria keep spill **default-OFF** (it needs a mandatory local scratch path), so the refusal branch is permanent in default configuration by design — which is why it must be detectable. The planner returns a bare `None` for both "no candidates" and "budget refused", so the caller logs one INFO line and the level stops consolidating permanently, indistinguishable from idle: segments constant, CPU ~0.4%, no error, no metric. Hit twice during this study.

## Harness fixes

- **base dimension** was hard-pinned to 128 in both harnesses — now a 2..4096 plausibility range, so beds over neural-embedding corpora work. Base/query agreement (the check that protects correctness) is unchanged.
- **checkpoint identity** included the snapshot `mtime_ns`, which is re-materialised every invocation — `--resume` could never match. Ships with a unit test.
- **metrics scrape** had a 30s timeout and no retry; it killed three large-collection bed builds.
- **ground truth** can now be generated exact-only, without a superset file.

## Commands run

```
python3 scripts/check_td_adr_unique.py      # 90 ADRs, 374 TD ids, 0 errors
python3 scripts/check_no_agent_attribution.py --range origin/develop..HEAD
python -m pytest tests/python/test_nprobe_sweep.py -q   # 6 passed
```

## Deployment impact

None. Documentation plus `scripts/bench/` and its test. No default is changed by this PR — TD-IVF-3 proposes one and gates it default-OFF for a follow-up.

### DIFF
--- a/docs/10-quality/td/TD-COMPACT-12-budget-refused-compaction-is-silent.adoc
+++ b/docs/10-quality/td/TD-COMPACT-12-budget-refused-compaction-is-silent.adoc
@@ -1,0 +1,109 @@
+// Copyright (C) 2026 ProximaDB
+// SPDX-License-Identifier: Apache-2.0
+= TD-COMPACT-12: a budget-refused compaction is indistinguishable from an idle level (INFO-only, no metric)
+:status: Open
+:dim: D3
+:pillar: OE
+
+[cols="1,3", options="header"]
+|===
+| Field | Value
+| Status | **Open**
+| Priority | **P2 — observability.** The underlying stall is TD-COMPACT-11 (P0). This entry is *only* about the operator being unable to detect it.
+| Component | `src/storage/common/compaction_memory.rs` :174–176 (planner returns `None`); `src/storage/common/compaction_utils.rs` :395 and :497 (the INFO log sites).
+| Evidence | Reproduced twice on 2026-08-08 during the coarse-PCA width study (TD-IVF-3): once at 128-d/3.3M and once at 768-d/491,655 rows. Details below.
+|===
+
+NOTE: This is **not** a duplicate of link:TD-COMPACT-11-deterministic-local-spill-compaction.adoc[TD-COMPACT-11].
+That entry covers the stall and builds the spill executor that resolves it. Its
+exit criteria explicitly keep spill **default-OFF** (spill demands a mandatory
+local scratch path, which not every deployment has). The refusal branch is
+therefore *permanent* in default configuration by design — which is exactly why
+it must be observable. TD-COMPACT-11 does not cover detection.
+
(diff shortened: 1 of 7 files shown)


vjsingh1984:
Rigorous review follow-up pushed in 1d9881228. It keeps the measured result but removes the unsupported dim-only production policy, makes TD-COMPACT-12 model typed/episode-based bounded-cardinality observability on the live retry path, scopes GET/query to hot-tier object-store request COGS, and adds an end-to-end exact-only ground-truth provenance test. Local verification: 10 focused Python tests passed; py_compile, diff check, TD/ADR uniqueness, and attribution guard passed. CI has been restarted by the push.

Issue #1542, opened 2026-08-07
records/scan cursor is minted against the collection object id but validated against the name — page 2 always 400s


vjsingh1984:
## Summary

A paginated `records/scan` cannot advance past the first page when the client addresses the collection **by name**. The cursor returned by page 1 is rejected by page 2:

```
page1: HTTP 200  records=1000  next_cursor=set
page2: HTTP 400  {"error":{"code":400,
        "message":"Invalid argument: scan cursor was issued for collection '2',
                   not 'retr_measure_proxima_codegraph_edges'",
        "type":"invalid_argument"}}
```

The cursor is stamped with the resolved numeric object id (`'2'`), and validated against the identifier in the request path (the name). Since the scan endpoint accepts a name and resolves it internally — `✅ Resolved 't' -> collection_id: '1'` appears in its own debug log — the two can never match for a name-addressed client.

Verified on `develop` @ `14860f1a7`, SST engine, 1,385 records, `limit: 1000`.

## Impact

Any collection larger than one page is effectively **unreadable in full** through the REST scan API unless the caller happens to know and use the internal object id. There is no documented way to obtain that id from the public API — `GET /collections` reports `collection_id`, but the scan path is documented and used by name.

For us this silently truncated a durability read: our client paginates, gets a 400 on page 2, and falls back to "no records", which looks exactly like an empty collection. (Our side degraded badly too — that is ours to fix — but the underlying cursor/identifier mismatch is here.)

## Expected

The cursor should round-trip against whatever identifier form the caller used, or the validation should compare **resolved** identities on both sides rather than the raw request string. Resolving the incoming identifier before the cursor check would fix it, since the resolution already happens a few lines earlier.

## Reproduction

1. Create a collection by name, insert > 1 page of records.
2. `POST /api/v2/collections/<name>/records/scan` with `{"limit": 1000}` → 200, `next_cursor` set.
3. Repeat with `{"limit": 1000, "cursor": <that cursor>}` → 400.

## Related

- The scan path also materializes the whole collection before filtering (TD-FPRUNE-2, #1533) — these are independent, but both make `records/scan` hard to use as a bulk read surface.

vjsingh1984:
This appears to be already fixed on `develop` by #1652, from the opposite direction to what was proposed here.

[...]

PR #1522, opened 2026-08-07
Promote develop → qa (2026-08-07): v0.3.0 (r3) — full develop HEAD

vjsingh1984:
Full develop HEAD frozen at `cbcb25875` (includes all 5 SDK test fixes: #1488 + #1518, and the recent dep bumps/refactors — nothing excluded, per 'bring all enhancements to main'). Supersede merge (qa tree superseded by develop; qa has no unique content). Runs the Python SDK tests at the qa gate (medium_tier). Any infra flake just needs a re-run; any real failure gets fixed on develop.

Next: qa → main (CI Success gate incl. python-test), delete stale April v0.3.0 tag, tag v0.3.0 + python-v0.3.0.

PR #1573, opened 2026-08-11
Promote develop → qa (2026-08-11): v0.3.0 (r5) — develop @ 9ac8432f9, supersedes r3

### LINKED ISSUE #1526: vector_count_threshold is accepted and documented but never read
> **This issue has been rewritten twice; see the comments.** The original filing claimed the SDK drops `memory_flush_size_mb` (wrong — my error) and then that no flush occurs until shutdown (also wrong — I did not wait past the 300s time floor). Both are retracted. What remains is the narrow finding below.

## Summary

`vector_count_threshold` is accepted in `[storage.wal_config]`, echoed back in the startup log, and never consulted by any flush decision.

## Trace

It has exactly one consumer — it is copied into the WAL performance config (`src/core/config.rs:1180`):

```rust
batch_threshold: self.vector_count_threshold,
```

`WalPerformanceConfig::batch_threshold` is then only ever **assigned**, never read:

- `write_ahead_log/config.rs:96` — field declaration
- `write_ahead_log/config.rs:158,440` — defaults / preset assignment
- `write_ahead_log/optimized_write_buffer_writer.rs:898` — assignment
- remaining hits are tests, or the unrelated `batch_threshold` on the columnar `hybrid_writer` config (a different struct that *is* read)

`FlushPolicy::from_performance` — the single flush decision (`flush_policy.rs`) — derives from `memory_flush_size_bytes`, `flush_interval_secs`, `wal_max_bytes`, the watermark fractions and `flush_floor_predicted_mb`. There is no record-count trigger at all.

## Why it is worth a small fix

Setting `vector_count_threshold = 500` and inserting 2,000 records does nothing, and the startup log actively confirms the value was read:

```
📋 Converting WALConfig from TOML: memory_flush_size_bytes=1048576 (1MB), vector_count_threshold=500
```

An operator reasonably concludes the knob is active. I did, and it sent me a long way down the wrong path.

## Suggested resolution

Whichever is cheaper is fine — this is a small honesty fix, not a feature request:

1. Remove the field (and its TOML documentation), since ADR-069's size/time/capacity triggers cover the intent; or
2. Wire a count trigger into `FlushPolicy` alongside the size target; or
3. Keep it and log a warning at startup that it is inert.

## Not part of this issue

Flushing itself works correctly. The ADR-069 time floor fires as designed:

```
✅ ADR-081 auto-flush [time] '1': 1000 vectors, 137862 bytes in 0.019s
```

and the size trigger is deliberately gated behind `flush_floor_predicted_mb` (TD-FLUSH-3) to avoid the tiny-segment pathology. Both behaved correctly under test on `develop` @ `b9d2c5d9c`.

vjsingh1984:
Supersedes **#1522** (r3), which pinned `87fca3544` from 2026-08-07 and had fallen **48 commits** behind develop. Cut as a fresh snapshot rather than back-syncing r3 — updating a promotion branch in place rewrites what was reviewed, and a snapshot keeps the promoted range auditable. Same pattern by which r4 superseded r3 earlier.

## What this promotes

`develop @ 9ac8432f9` → `qa @ 1897e80c4` — **377 commits**:

| type | count |
|---|---|
| feat | 135 |
| docs | 72 |
| refactor | 68 |
| fix | 58 |
| chore | 16 |
| test | 15 |
| ci | 10 |
| bench / build | 3 |

The 44 commits qa carries that develop does not are the accumulated merge commits from previous promotions — qa holds no unique work, so this is a clean forward promotion, not a divergence to reconcile.

## Why the head is a pinned branch, not `develop`

`develop` is moving actively (it advanced twice during the few minutes spent preparing this). A promotion PR whose head is `develop` re-runs the full heavy gate on every subsequent merge to develop; pinning a snapshot SHA keeps this PR's gate result meaningful and stops it churning underneath reviewers.

If the gate demands a newer base, cut **r6** at that point rather than force-updating this branch.

## Included fixes worth calling out

Among the 377 are several from the recent Victor ↔ ProximaDB co-design work:

- **#1557** — retire the inert `vector_count_threshold` knob (closes #1526)
- **#1525** — send the configured storage engine on collection create (embedded collections had all been HELIX regardless of config)
- **#1528** — surface segment-listing failures instead of reporting an empty collection
- **#1531** — repair the `python` feature build; the `proximadb_embedded` wheel was unpublishable
- **#1533** — TD-FPRUNE-2: `records/scan` materializes the whole collection before filtering
- **#1523** — TD-FLUSH-8 catalog-authoritative collection resolution at the auto-flush boundary

## Still open against this line

- **#1524** — ORION persists edges but cannot read them back after a restart. Re-diagnosed as a *read* bug, not a durability one; silently degrades `/graphs/{id}/rag` and the graph arm of hybrid fusion.
- **#1527** — `record_count` reports 0 for a populated collection.
- **#1542** — `records/scan` cursor is minted against the collection object id but validated against the name, so page 2 always 400s for a name-addressed client.

None are regressions introduced here; all predate this range.


vjsingh1984:
## Conflicts resolved, and the cause fixed

**Before:** 47 conflicts. **Now:** `MERGEABLE`, and the promoted tree is byte-exact `develop`.

### Why there were conflicts at all

Not drift on `develop` — its CI gates every merge and it is consistent. The promotions themselves are **squash-merged**:

```
1897e80c4 parents=1   Promote develop → qa (2026-08-05)
69b3ad5c9 parents=1   Promote develop → qa (2026-08-01)
a30125c28 parents=1   Promote develop → qa (2026-07-28)
```

A promotion branch normally merges the target in to resolve conflicts — but squashing the PR **discards that merge**, so `qa` receives develop's *content* without its *ancestry*. The merge-base between the branches then stops advancing:

```
merge-base:            1ab38bac7   2026-07-21   (the last real merge)
develop ahead of it:   377 commits
```

Git re-diffed that entire range on every promotion. The 47 conflicts were almost all phantom, and each cycle compounds.

### How it was resolved

`git merge -s ours origin/qa` — keeps develop's tree byte-exact while recording `qa` as a second parent. Safe here because **`qa` carries no unique work**, verified rather than assumed:

- its only non-promotion commits are CI fixes (`swap headroom`, `CARGO_BUILD_JOBS=1`, `cancel-in-progress` exemption) — all already on develop, identical counts, and `qa-gate.yml` was not even among the conflicts;
- the 34 files "only on qa" are pre-decomposition paths that develop relocated into crates (TD-DECOMP-4x);
- develop's CI actions are strictly newer (`checkout@v7`/`cache@v6` vs `v4`/`v0.0.10`).

Safety gate: `git diff origin/develop HEAD` is **empty**. Ancestry check: develop commits outside the merge-base went **377 → 0**.

### Guard added

`promotion-ancestry-guard` in `qa-gate.yml`, on PRs targeting `qa`/`main`. It asserts the exact invariant — the merge-base must **be** the base tip — rather than a tuned threshold.

Calibrated against real history: branches that had merged the target in measure `0`; this branch, cut as a bare develop snapshot, measured `44`. **My first draft allowed 150 and would have passed the broken case**, which is why it now asserts the invariant instead of guessing a limit.

### ⚠️ This PR must be merged with a MERGE COMMIT, not squashed

Squashing discards the `-s ours` merge and puts the next promotion straight back to a severed merge-base. The whole point of the second parent is that r6 should then open with **zero** conflicts — that is the test of whether this actually worked.

PR #1610, opened 2026-08-12
chore(release): back-merge main into qa (align history before v0.3.0 promote)

vjsingh1984:
Reconcile qa's history with main so the v0.3.0 qa→main promote (#1594) satisfies main's strict up-to-date gate + the new promotion-ancestry-guard (#1590).

Rebuilt on current qa (c4d7c73f0) after the r6 develop→qa promote (#1592) advanced qa. ORT(qa, main) == qa exactly (0 conflicts, tree byte-identical); ancestry-only. Contains qa's tip (satisfies qa's strict gate) and merges main in.

Per docs/04-operations/release/PROMOTION_PROCEDURE.adoc: **merge commit, never squash.**


vjsingh1984:
Redundant: #1594 already merged qa→main (main now contains qa). The version-sync fix (#1622) will reach qa via the next develop→qa promote from develop, which already contains main (#1603) — so that promote restores qa's main-ancestry naturally per PROMOTION_PROCEDURE.adoc. No separate reconcile needed.

